PR #420, opened 2019-08-28
Handling api token in more secure way

### LINKED ISSUE #240: Token for API
[A token is used for the API. ](http://oep-data-interface.readthedocs.io/en/latest/api/how_to.html#authenticate)

How do we handle/store the token in a secure way? We definitely should not hard code it in our python scripts and push it to GitHub. 
A small search gives us this options. Feel free to add other or comment:

- [**stups-tokens**](https://pypi.python.org/pypi/stups-tokens)
- store it in a **config file**, similar to [oemof.db](https://github.com/oemof/oemof.db)
- [**keyring**](https://pypi.python.org/pypi/keyring)

Bachibouzouk:
Closes #240 

The user has the option of storing the token in an environment variable or in the config file directly.

### DIFF
--- a/api/tutorials/token_config.py
+++ b/api/tutorials/token_config.py
@@ -1,0 +1,26 @@
+from subprocess import run
+# untrack the change to this file (prevent pushing token to github)
+run(['git', 'update-index', '--assume-unchanged', __file__])
+
+oep_url = 'http://oep.iks.cs.ovgu.de/'
+# Paste your token here
+oep_token = ''
+
+
+def get_oep_token():
+    """
+    Function which will warn the user if the oep token is empty
+    :return oep token of user:
+    """
+    if not oep_token:
+        raise(
+            Warning(
+                'The OEP token you provided in the api/token_config.py file is empty.\n'
+                'Please sign in to https://openenergy-platform.org/user/login/?next=/ '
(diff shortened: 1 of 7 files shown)


api/tutorials/token_config.py
```diff
@@ -0,0 +1,26 @@
+from subprocess import run
+# untrack the change to this file (prevent pushing token to github)
+run(['git', 'update-index', '--assume-unchanged', __file__])
```
Bachibouzouk: This line will prevent the changes to this file to be tracked by git upon the first time a user does `from token_config import get_oep_token()`.

As experienced user will prefer the environment variable option, I assumed that the chances a user commit and push their change after copy-pasting their token in the file to be pretty slim.

christian-rli:
Nice trick! I tried it out and it checks out.

Bachibouzouk:
Learned that these tutorial belong to the [example repo](https://github.com/OpenEnergyPlatform/examples), so I moved the commits with `cherry-pick` to https://github.com/OpenEnergyPlatform/examples/pull/61

Bachibouzouk:
Relocated to https://github.com/OpenEnergyPlatform/examples/pull/61

PR #425, opened 2019-08-30
Update egoio in requirements

### LINKED ISSUE #421: Egoio is not installed via pypi
This is because the current release cannot be installed, it should be changed once https://github.com/openego/ego.io/issues/78 is closed

Bachibouzouk:
Fixes #421 

The latest [release](https://github.com/openego/ego.io/pull/82) of egoio fixed the error preventing its pip install and added geoalchemy in its requirements. Therefore we can again simply add `egoio` in the requirements.txt


Bachibouzouk:
Done in #428

PR #428, opened 2019-09-09
Update psycopg in requirements.txt

### LINKED ISSUE #421: Egoio is not installed via pypi

Bachibouzouk:
Closes #421 

Install `egoio` via pip and choose `psycopg` over `psycopg-binary` (after discussion in https://github.com/OpenEnergyPlatform/oedialect/issues/15)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ bibtexparser>=0.6.2
 django-colorfield>=0.1.12
 requests>=2.13.0
 geoalchemy2<=0.4.1,>=0.3.0  # because egoio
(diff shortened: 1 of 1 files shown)


Bachibouzouk:
At the moment psycopg2-binary is [installed](https://github.com/openego/ego.io/blob/d7a6d16b0ea03b942890140236bf2e3fab579dbe/setup.py#L19) by egoio `setup.py`. Version higher than `2.7.6` will not work (display of tables fail in API)

PR #429, opened 2019-09-11
Fix API bugs

MGlauer:
* Fix typing issues with different date/time formats
* Implement type specification in value clauses
* Fix rounding issues with `float`

### DIFF
--- a/api/actions.py
+++ b/api/actions.py
@@ -34,7 +34,7 @@ from api.sessions import (
 )
 from dataedit.models import Table as DBTable
 from dataedit.structures import MetaSearch
-from oeplatform.securitysettings import PLAYGROUNDS
+from oeplatform.securitysettings import PLAYGROUNDS, UNVERSIONED_SCHEMAS
 
 pgsql_qualifier = re.compile(r"^[\w\d_\.]+$")
 
@@ -69,7 +69,7 @@ def get_table_name(schema, table, restrict_schemas=True):
     if schema.startswith("_") or schema == "public" or schema is None:
         raise PermissionDenied
     if restrict_schemas:
-        if schema not in ["model_draft", "sandbox", "test"]:
+        if schema not in PLAYGROUNDS and schema not in UNVERSIONED_SCHEMAS:
             raise PermissionDenied
     return schema, table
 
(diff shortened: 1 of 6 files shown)


api/actions.py
```diff
@@ -69,7 +69,7 @@ def get_table_name(schema, table, restrict_schemas=True):
     if schema.startswith("_") or schema == "public" or schema is None:
         raise PermissionDenied
     if restrict_schemas:
-        if schema not in ["model_draft", "sandbox", "test"]:
+        if schema not in PLAYGROUNDS and schema not in UNVERSIONED_SCHEMAS:
```
Bachibouzouk:
```suggestion
        if schema not in PLAYGROUNDS + UNVERSIONED_SCHEMAS:
```

api/actions.py
```diff
@@ -1090,7 +1020,7 @@ def data_delete(request, context=None):
     setter = []
     cursor = load_cursor_from_context(context)
     result = __change_rows(request, context, target_table, setter, ["id"])
-    if orig_schema in PLAYGROUNDS:
+    if orig_schema in PLAYGROUNDS or orig_schema in UNVERSIONED_SCHEMAS:
```
Bachibouzouk:
```suggestion
    if orig_schema in PLAYGROUNDS + UNVERSIONED_SCHEMAS:
```

api/actions.py
```diff
@@ -1116,7 +1046,7 @@ def data_update(request, context=None):
         setter = dict(zip(field_names, setter))
     cursor = load_cursor_from_context(context)
     result = __change_rows(request, context, target_table, setter)
-    if orig_schema in PLAYGROUNDS:
+    if orig_schema in PLAYGROUNDS or orig_schema in UNVERSIONED_SCHEMAS:
```
Bachibouzouk:
```suggestion
    if orig_schema in PLAYGROUNDS + UNVERSIONED_SCHEMAS:
```

api/actions.py
MGlauer: I am not sure which version I prefer. The first version performs better for large lists. Yours is undoubtedly more readable  and - according to `timeit` - more efficient for small lists. Interesting. As I expect small sets of schemas, I will go with your version.

PR #432, opened 2019-09-16
Fix/factsheet remodel

### LINKED ISSUE #36: [Factsheets] Fact sheets: References
Citation reference and References to reports produced using the model
- larger text fields
- plus button for more than one reference, opening an additional text field
- Citation DOI fields next to each field

Citation reference info text
"Publication(s) to be referenced when citing the model; e.g. Doe, John (2014), An electricity market model for Europe, Journal of Modelling Sciences, Vol 10, Issue 80."

Larger text fields also for all other points.

Larger scale usage
- MF: rename to "Model usage"; info text: Who uses the model?
- FF: rename to "Framework" usage; info text: Who uses the framework?

validation
options: peer-reviewed, outcomes cross checked with other models or data, other + text

### LINKED ISSUE #38: [Factsheets] fact sheets: several display errors
At different points in the sections "Coverage" and "Mathematical properties" the submitted fact sheet show "True" or an array of the word "True" instead of a check mark or a list of the selected options.

The label "framework yes text" should not appear in the fact sheet. If there is a text it should stand for itself (or give it a more beautiful name like name of the underlying framework). If there is no text, this line should not be part of the fact sheet.

The points "planned to open up in the future" and "costs" appear in the ready fact sheet of open source models, but can not be filled out in the form.

The "license other text" should not appear in the final fact sheet or it should be renamed. When choosing a licence from the list after choosing the "other option" and entering a name in the other field first, in the form the "other option" and the entered name disappear, but in the final fact sheet the name or the standard text "not jet decided" is displayed nevertheless and the selection is ignored (model and framework fact sheet).

As above the label "external optimizer yes text" should not appear.

Do we prefer a check box with yes-text for "Additional software"? Preventing people to write none / no etc. The field should not be mandatory (please, change that also in the framework fact sheet).

### LINKED ISSUE #50: [Factsheets] MF openness, source code available
open source = yes -> source code available = yes:
- GitHub checkbox appears
- Rename "link to source code" as "Access to source code", Info text: "Where is the source code available (e.g. link, email address)?"; type: not link, but text field.

### LINKED ISSUE #74: [Factsheets] make links in fact sheets clickable

### LINKED ISSUE #82: [Factsheets] SF Ask for contact information
@ludee Do we want to ask for contact information in the Scenario Fact Sheet?

### LINKED ISSUE #84: [Factsheets] MF - Coverage - final / end energy
Taking the fact,  that "electricity, heating, mobility and oil" are not really sectors (see openmod wiki fact sheet) as a starting point we defined "Modeled demand sectors" (Households, Industry, Commercial sector, Transport) and "Modeled energy sectors (final energy)" (Electricity, Heat, Liquid fuels, Gas, Others) as well as "Modeled energy carriers (primary energy carriers)" (Gas, Liquids, Solid, Renewables with options each). Furthermore, there is "Modeled technologies: components for generation or conversion" (Renewables, Conventional with options each).
See: http://193.175.187.164/factsheets/models/add/

That seems to be to much.
1. Please, make the option at the third level to voluntary options, that appear when the second level is selected.
2. @ludee: How do we re-organise the questions above? Can we just leave out "Modeled energy carriers (primary energy carriers)" and add the options "Geothermal heat" (RE), "Hydrogen" (Conv. + RE), "Biofuels" (RE) to Modeled technologies: components for generation or conversion"?

### LINKED ISSUE #85: [Factsheets] MF - Coverage - Properties electrical grid
The options should be:
AC load flow model
DC load flow model
transshipment model (instead of Net transfer capacities)
single-node / copper plate model
other

### LINKED ISSUE #311: [Factsheets] add new factsheet model questions
**General Information Question 4**
-Institutions is listed 2 times 

**subregister  "Model Integration"**
-Input/output data just one selection possible sometimes more then one data format is given (checkbox instead of drop down?)

### LINKED ISSUE #401: Scenario factsheet: some selected technologies onyl displayed as "true"
Not all technologies clicked are depicted as such, but some only as "true"

grafik

solar-c:
Closes #311, closes #401, closes #74, closes #36, closes #38, closes #50, closes #82, closes #84, closes #85


4lm:
Hi @solar-c,

I'm reviewing your pull request and have a few remarks:

#311: In https://github.com/OpenEnergyPlatform/oeplatform/pull/432/commits/57a35c46a427fbab35fd11965450930b6e49d126 you introduced a method `create_devuser()` in `login/models.py`, which is commented out right in the next commit https://github.com/OpenEnergyPlatform/oeplatform/pull/432/commits/c9ad6edc1d84193319a6c2a88ce73a1a74b731bb. In know that you use this method for easily creating a user while developing locally. IMO it would be OK to introduce this method and remove the multline comment around it.

#401: good!

#74: good! Maybe using a [regex](https://stackoverflow.com/questions/3809401/what-is-a-good-regular-expression-to-match-a-url) or a more complex condition would be a nice to have in the future to exclude false positives. But IMO not a must for now.

#36: This seems to have a bug (maybe I'm wrong). For reproduction, I create a model and framework factsheet and go to "References" ...

... in model factsheet:

- the name "Model usage" is correct, but the description is wrong "Is this framework ...", should be "Is this model ..." - right?

... in framework factsheet:

- the description is correct "Is this framework ...", but the name "Model usage" is wrong, should be "Framework usage" - right?


#38: good!

#50: good!

#82: good!

#84: good!

#85: good!

modelview/models.py
```diff
@@ -674,6 +674,11 @@ def __str__(self):
         help_text="Who are the authors of the study and for which institution do they work?",
         max_length=1000,
     )
+    contact_email = EmailField(
+        verbose_name="Contact (e-mail)",
+        help_text="Please provide the mailadress of the contact person.",
```
Bachibouzouk: email address

requirements.txt
```diff
@@ -19,3 +19,4 @@ django-axes==2.3.3
 shapely>=1.5.17
 markdown2
 alembic
+github+https://github.com/solar-c/django-bootstrap3.git
```
Bachibouzouk: Is it a fork? We should maybe host this code also on the OEP github account

modelview/static/css/form-style.css
```diff
@@ -57,3 +57,7 @@ required:0;
     cursor: pointer;
     color: red;
 }
+
+.modeledTec th {
```
Bachibouzouk: Update is not a very descriptive commit name ;)

login/models.py
```diff
@@ -51,6 +51,17 @@ def create_superuser(self, name, email, affiliation):
...
+        if not email:
+            raise ValueError("An email address must be entered")
+        if not name:
+            raise ValueError("A name must be entered")
+        user = self.model(
+            name=name, email=self.normalize_email(email), affiliation=name, is_mail_verified=True
+            )
+        user.save(using=self._db)
+        return user"""
+
```
Bachibouzouk: I agree with @4lm with leaving the code without the quotes

requirements.txt
solar-c: yes, its a fork. we could also fork it to the OEP github. but lets do it in a new branch/pull request.

solar-c:
#311: removed multiline comment around new method create_devuser()

#74: maybe if we have some time left :) good remark!

#36: fixed it

Everything should be resolved now

PR #441, opened 2019-10-07
Feature/metadata v1 4

### LINKED ISSUE #427: Missing OEP-Metadata v1.4 Support
Uploading metadata in version 1.4.0 to the OEDB won't result in a correct display thereof. Only uploading a comment in version 1.3 will be rendered correctly. This was mentioned in a few other issues already (https://github.com/OpenEnergyPlatform/data-preprocessing/issues/26 https://github.com/OpenEnergyPlatform/data-preprocessing/issues/27), but did not get its own issue as far as I can see.

Bachibouzouk:
Closes #427 

### TODOS

- [x] Fix metadata representation issues: https://github.com/OpenEnergyPlatform/oeplatform/pull/441#issuecomment-539534258 - @4lm
    - [x] Have a function which loops automatically through the structure of the metadata (6222b8cd686c0f3b3c1a83db1bdbbcc5ede08bfa)
    - [x] The element within lists have to be created with a js [function](https://github.com/OpenEnergyPlatform/oeplatform/blob/326e97c30eaddd987a3fe48e8ea850fbe204f6c6/dataedit/templates/dataedit/list_field.html#L18) called within [dataedit/templates/dataedit/list_field.html](https://github.com/OpenEnergyPlatform/oeplatform/blob/develop/dataedit/templates/dataedit/list_field.html#L18)
    - [x] Figure out how one can add the js function described above for the new fields (e44ffc567f3e5d409e579d3de43291a8ffad91ee)
- [ ]  Manage metadata versions with OMI - @4lm 
- [x]  Manage POST request - @Bachibouzouk 
    - [x] Assign the values from the query dict sent by the post request to the metadata template (933645cc0ae31e0c0aa0ac9b7cbc3a6971bbd4c0)

### Resources

- https://github.com/OpenEnergyPlatform/examples/wiki/Metadata-Description
- https://github.com/OpenEnergyPlatform/omi


4lm:
Hi @Bachibouzouk,

I had a look at v1.3 and v1.4 in the datedit view of the app and compared it with the JSON they originate from and found following points:

In 1.3:

- Sources -> URLs missing
- License -> URLs not clickable
- Metadata_version has underscore
- Resources -> Fields. Fields are not showing up properly under Resources (scrambled), but also a section Fields (not in JSON), there they are rendered properly

In 1.4:In 1.3:

- Sources -> URLs missing
- License -> URLs not clickable
- Metadata_version has underscore
- Resources -> Fields. Fields are not showing up properly under Resources (scrambled), but also a section Fields (not in JSON), there they are rendered properly

In 1.4:

- Id -> URL not clickable, general issue I stop listing it from now on ...
- Publicationdate not formatted properly
- Context -> Sourcecode, Fundingagency, Fundingagencylogo, publisherLogo not rendered properly, doesn't take camelCase naming into account
- Temporal -> Aggregationtype (see Context, general issue I stop listing it from now on ...
- Sources -> License, "title" is but "name" (e.g. "CC0-1.0") is not rendered
- Licenses, "title" is but "name" (e.g. "CC0-1.0") is not rendered
- Resources -> Schema-> Fields. see 1.3
- Resources -> Schema -> Foreign Keys. Is "info There is no valid entry for this field" correct?
- _comment: Why the underscore?


- Id -> URL not clickable, general issue I stop listing it from now on ...
- Publicationdate not formatted properly
- Context -> Sourcecode, Fundingagency, Fundingagencylogo, publisherLogo not rendered properly, doesn't take camelCase naming into account
- Temporal -> Aggregationtype (see Context, general issue I stop listing it from now on ...
- Sources -> License, "title" is but "name" (e.g. "CC0-1.0") is not rendered
- Licenses, "title" is but "name" (e.g. "CC0-1.0") is not rendered
- Resources -> Schema-> Fields. see 1.3
- Resources -> Schema -> Foreign Keys. Is "info There is no valid entry for this field" correct?
- _comment: Why the underscore?

So before I  implement the metadata management with OMI, I will try to fix those issues first ...

4lm:
@Bachibouzouk. Please have a look at https://github.com/OpenEnergyPlatform/oeplatform/pull/441/commits/e1b0ce68419c3c7446dd433b5cb91a0c7867b323. What do you think, shall I put the regex into a constants python file/module, so that we can use it in the whole project? If so, where would you put it? If so, maybe we also have other things we could check with further regular expressions and which also could life in this location?

I also add @solar-c to this (FYI), because I know that she had to check for URLs in one of here last pull requests ...

Bachibouzouk:
>     * _comment: Why the underscore?

I think some of the fields are not meant to be displayed in public. I would guess that `resources`, and `_comment` are not meant to be displayed "as is". I also guess `context` should be rendered in a special way (ie display the logo if there is one)

Bachibouzouk:
>     * Licenses, "title" is but "name" (e.g. "CC0-1.0") is not rendered

@4lm - `title` is the human readable name whereas `name` is not used for display purposes (as far as I understood). [Here](https://github.com/OpenEnergyPlatform/examples/wiki/Metadata-Description) is a description of what the different fields mean (sorry for not sharing that beforehand)

Bachibouzouk:
> What do you think, shall I put the regex into a constants python file/module, so that we can use it in the whole project?

Good idea, I would put it in a folder named for example `utils` in the root of the repository

Bachibouzouk:
@MGlauer, @Ludee - how did you think omi would be used to manage the different versions of metadata on the OEP? I am interested to know what is the plan here before I do too much work for nothing because it is already done somewhere else.

@MGlauer - when I change things in `dataedit/static/dataedit/metadata.js`, I don't see those changes reflected when I run the server locally (even if I run `python manage.py collectstatic`), is there an additional command I should run?

MGlauer:
In order to include a different version of the metadata, you just need a different [parser](https://github.com/OpenEnergyPlatform/omi/blob/dev/src/omi/dialects/base/parser.py). The problem is that old metadata is often incomplete or even not well-formed.

MGlauer:
> @MGlauer - when I change things in `dataedit/static/dataedit/metadata.js`, I don't see those changes reflected when I run the server locally (even if I run `python manage.py collectstatic`), is there an additional command I should run?

This might be an issue caused by your browser. Some browsers try to be smart and use cached statics instead of reloading them. Try to deactivate the caching

Bachibouzouk:
> > @MGlauer - when I change things in `dataedit/static/dataedit/metadata.js`, I don't see those changes reflected when I run the server locally (even if I run `python manage.py collectstatic`), is there an additional command I should run?
> 
> This might be an issue caused by your browser. Some browsers try to be smart and use cached statics instead of reloading them. Try to deactivate the caching
> 
> PS: `collectstatic` should not be required if you run Django's development webserver (`manage.py runserver`)

@MGlauer - The cache is not cleared if I stop the server and I close my browser? I will try your solution, thanks :)

dataedit/metadata/__init__.py
```diff
@@ -1,153 +1,237 @@
 from api import actions
-from dataedit.metadata import v1_3 as __LATEST
+from dataedit.metadata import v1_4 as __LATEST
+
+from dataedit.metadata.v1_3 import TEMPLATE_v1_3
+from dataedit.metadata.v1_4 import TEMPLATE_V1_4
```
Bachibouzouk: This will be changed after integration of OMI

dataedit/metadata/v1_3.py
```diff
@@ -3,6 +3,66 @@
 from . import v1_2
```
Bachibouzouk: This template should be loaded from OMI

dataedit/metadata/v1_4.py
```diff
@@ -0,0 +1,139 @@
+from api import actions
+
+TEMPLATE_V1_4 = {
```
Bachibouzouk: this template should be loaded from OMI in the future

dataedit/metadata/widget.py
```diff
@@ -0,0 +1,247 @@
...
+
+from django.utils.safestring import mark_safe
+
+from dataedit.metadata import METADATA_HIDDEN_FIELDS
+
+LICENSE_KEY = 'license'
+COLUMNS_KEY = 'fields'
+
+
+class MetaDataWidget:
```
Bachibouzouk: This create either an html normal view or an html form view of a nested structure which can contain dict, list and strings

dataedit/static/dataedit/metadata.js
```diff
@@ -1,139 +1,119 @@
...
-    prefix='temporal';
-    indexes[prefix] = indexes[prefix] +1;
-    var $container = create_box($parent);
 
-    if('undefined' === typeof obj){
-        var obj={};
-        obj.start='';
-        obj.end='';
-		obj.resolution='';
+function add_list_objects(prefix){
```
Bachibouzouk: this function replaces all the add_* functions, it copies an object from the list, re-initializes its values and forms and replace the ids of the html component with a new one

dataedit/templates/dataedit/dataview.html
```diff
@@ -155,126 +155,18 @@ <h4 class="panel-title">
...
-                            No description
-                        {% endif %}
-                        <hr>
-                    {% endfor %}
-            </td>
-        </tr>
-        {% endif %}
-    </table>
+     <h4>Metadata</h4><a href="{{table}}/meta_edit"> Edit</a>
+        {{meta_widget}}
```
Bachibouzouk: The render function of the widget automatically generated the code (and can do so for a future updated version of metadata)

Bachibouzouk:
@4lm - let's postpone the integration with OMI to a subsequent PR, this one is already consequent!

4lm:
@Bachibouzouk, mhh, if I try to edit v1.4 metadata I get an error. I will look into it ...

4lm:
Fixed it https://github.com/OpenEnergyPlatform/oeplatform/pull/441/commits/0c7e72bc665b0ea982e87929e3300fb7c1e42865, none type `None` was not handled.

PR #434, opened 2019-09-29
WIP: fix/design rework welcome page

johannwagner:
I implemented the first part of the mockup. 
I have to contact Bryan about some things, I will report feedback at this point.

localhost_8000_(iPad Pro)

PR #447, opened 2019-10-21
Feature/modularize templating

4lm:
image


4lm:
Hi,

please review this pull request! I (i) modularized the templating (also compare with photo above), (ii) introduced a naming convention for the blocks and (iii) introduced as JavaScript section add the bottom of the body.

Now every section (header, left-sidebar (new), main, right-sidebar, footer) is wrapped by a block. I also kept the old sub-section block structure, to not destroy the already given layout.

literature/templates/literature/base.html
```diff
@@ -1,5 +1,5 @@
 {% extends "base/base.html" %}
 
-{% block content %}
+{% block main-content-body %}
 
-{% endblock %}
\ No newline at end of file
+{% endblock %}
```
Bachibouzouk:
```suggestion
{% endblock main-content-body%}
```

modelview/templates/modelview/base.html
```diff
@@ -22,7 +22,7 @@
 want to compare. OEP will give us tables (views) that facilitate the comparison.
 <hr>
 {% endblock %}
```
Bachibouzouk:
```suggestion
{% endblock main-right-sidebar-content %}
```

base/templates/base/base.html
```diff
@@ -1,129 +1,131 @@
...
-        <link rel="stylesheet" href="{% static 'css/base-style.css' %}">
-        {% block after-header %}{% endblock %}
-    </head>
-
+<html lang="en">
+<head>
+    {% block before-head %}{% endblock before-head %}
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
```
Bachibouzouk: Can you describe why meta is useful here? (just for documenting the steps and to ease the review process of a file with many block of code changes)

literature/templates/literature/base.html
4lm: Right now, 

> Thanks for these improvements suggestions @4lm, could you perhaps merge release/v0.5.0 locally to test that this is ok with the design there?
> 
> Could you summarize which are the lines in `base.html` where you added/removed something by commenting on the file in github?

Yes, will do!

modelview/templates/modelview/base.html
4lm: >`- {% endblock %}`
>`+ {% endblock main-right-sidebar-content %}`

@Bachibouzouk. Right now, only endblocks in `base/templates/base/base.html` have a more verbose naming, because, I would have to update all endblocks by hand with the more verbose naming. I can do that, if you agree with the more verbose naming. Personally if the code blocks are larger this helps my a lot, but can be an overkill if the code blocks are small. On the other hand IMO it makes sense to have everywhere the same... What do you think? Do you like this more verbose naming? Shall we implement it all templates?

base/templates/base/base.html
4lm: @Bachibouzouk, this is the minimal meta set, that nearly always should be implemented.

- `<meta charset="UTF-8">`: charset info
- `<meta name="viewport" content="width=device-width, initial-scale=1.0">`: viewport (import as responsive design base): https://www.w3schools.com/css/css_rwd_viewport.asp
- `<meta http-equiv="X-UA-Compatible" content="ie=edge">`: compatibility info for stating, that we only support the Microsoft edge browser (which soonish uses chromium as base :rocket:)

4lm:
> Thanks for these improvements suggestions @4lm, could you perhaps merge release/v0.5.0 locally to test that this is ok with the design there?

Makes IMO little sense until this isn't solved: #448. But I don't think it will be a problem. I can manually solve the merge conflicts, which will be there, but should be no problem, because I do not destroy the already given layout structure ...

base/templates/base/base.html
```diff
@@ -1,129 +1,131 @@
+{% load static %}
+{% load bootstrap3 %}
```
4lm: - I always put the tags for loading of static files, boostrap etc. on top of template file (for having a better overview), that's why I factored it out of the head.
- I also used the more common {% load static %} instead of {% load staticfiles %}, because we don't host static files on external services - right?: https://stackoverflow.com/questions/24238496/what-is-the-difference-between-load-staticfiles-and-load-static

base/templates/base/base.html
```diff
@@ -1,129 +1,131 @@
...
-        <link rel="shortcut icon" href="{% static 'favicon.ico' %}"  type="image/x-icon" />
-        <title>
-            OEP{% block title %}{% endblock %}
-        </title>
-        <link rel="stylesheet" href="{% static 'css/theme-style.css' %}">
-        <link rel="stylesheet" href="{% static 'css/base-style.css' %}">
-        {% block after-header %}{% endblock %}
-    </head>
-
+<html lang="en">
```
4lm: I added `lang="en"` for defining the used language, this makes sense if we have one primary language.

If we go one day i18n, this should be changed, but should be done by either using subdomains "`en.|de.|fr.|...`" or "`domainname/de-DE|domainname/fr-FR|domainname/en-US|...`". Because search engines don't like websites that change their content (language texts) on the same URL magically ...

base/templates/base/base.html
4lm: Please see: https://github.com/OpenEnergyPlatform/oeplatform/pull/447/files#r337005466

base/templates/base/base.html
```diff
@@ -1,129 +1,131 @@
...
+<head>
+    {% block before-head %}{% endblock before-head %}
```
4lm: I did put this on top of the head section.

base/templates/base/base.html
```diff
@@ -1,129 +1,131 @@
...
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>OEP{% block title %}{% endblock title %}</title>
+    <link rel="shortcut icon" href="{% static 'favicon.ico' %}"  type="image/x-icon" />
+    {% bootstrap_css %}
+    <link rel="stylesheet" href="{% static 'css/theme-style.css' %}">
+    <link rel="stylesheet" href="{% static 'css/base-style.css' %}">
```
4lm: I removed the JavaScript in this section and added it to a section a the bottom of the body. JS should be always a the bottom of the body section for not blocking HTML rendering ...

base/templates/base/base.html
```diff
@@ -1,129 +1,131 @@
...
+    {% block after-head %}{% endblock after-head %}
+</head>
 <body>
```
4lm: My block structure changes are mainly in the following `<body>` section. I divided the block structure into sections and sub-sections:

1.) The main sections are:

- header (formerly named "nav")
- main
- footer

2.) I kept the old block layout in sub-blocks and introduced a new naming convention for simplifying things, e.g.:

- main
    - main-content-body (formerly just called "content")

3.) I also introduced an additional left sidebar to the given right sidebar, which both are part of the main section, to not destroy the old layout structure (normally I would put the sidebars into an independent section):

- main
    - main-right-side-bar
    ...
    - main-left-side-bar

base/templates/base/base.html
```diff
@@ -1,129 +1,131 @@
...
+            </div>
+    </nav>
+{% endblock footer %}
 
+{% block body-bottom-js %}
+    {% block before-body-bottom-js %}{% endblock before-body-bottom-js %}
+    {% bootstrap_javascript jquery=1%}
+    <script type="text/javascript" src="{% static '/js/jsi18n'%}"></script>
+    {% block after-body-bottom-js %}{% endblock after-body-bottom-js %}
+{% endblock body-bottom-js %}
```
4lm: Javascript section at the bottom with subsections. Normally only block `after-body-bottom-js` should be overwritten.

4lm:
> > Thanks for these improvements suggestions @4lm, could you perhaps merge release/v0.5.0 locally to test that this is ok with the design there?
> 
> Makes IMO little sense until this isn't solved: #448. But I don't think it will be a problem. I can manually solve the merge conflicts, which will be there, but should be no problem, because I do not destroy the already given layout structure ...

#448 is solved and the result is, that `fix/design-rework-welcome-page` is the correct branch to merge. Will do and report back!

4lm:
> #448 is solved and the result is, that `fix/design-rework-welcome-page` is the correct branch to merge. Will do and report back!

Aborted the merge, because the code got to compicated, this needs further discussion, here: https://github.com/OpenEnergyPlatform/oeplatform/pull/434#issuecomment-545017454

johannwagner:
I implemented the changes of this branch manually at #458. 
I propose to close this PR, because it is unmergable.

Bachibouzouk:
@4lm can you have a look at #458 ?

4lm:
> @4lm can you have a look at #458 ?

Yes, will do

PR #449, opened 2019-10-21
Feature/improve meta edit view ux

4lm:
work in progress ...


Bachibouzouk:
@4lm - is this to be transferred on the latest version of  `develop`?

jh-RLI:
This is outdated since we introduced the metadata wizard. Do you agree? @wingechr

wingechr:
> This is outdated since we introduced the metadata wizard. Do you agree? @wingechr

yes.

PR #450, opened 2019-10-22
Feature/master merge

MGlauer:
I merged the master branch into the develop branch. Merging this into the develop will have two implications:
* The new (adapted) redesign will be available on develop 
* BUT: The original redesign will be hard to merge into dev after that due to the cherry-pick. I will try to restore the commit tree structure in a way that makes it easier, though.


Bachibouzouk:
@MGlauer : I've discussed with @4lm and we think we should not mix the histories of master with develop (ie push always go from develop to master, never in reverse). What could be a solution, would be to keep working on the new design in #434 and then merge this in develop. Then, before we do a release to master we do a rebase of master to remove the hotfixed commits from release/v0.5.0 and immediately after we push the clean release.

MGlauer:
> I've discussed with @4lm and we think we should not mix the histories of master with develop (ie push always go from develop to master, never in reverse).

@Bachibouzouk Can you give a rationale for that? What I did is what the [Driessen workflow](https://nvie.com/posts/a-successful-git-branching-model/) proposes but instead of merging the release-branch back at release, I merged the master which is at the same commit. So those are equivalent. I can rename the commit if you want. The problem is, that there are commits on master that are not on dev, which should not be the case and would cause massive problems in the long run. Hence this PR.

As @gnn and I understand it, your approach would mean that we rewrite the history of a pubic branch, which will break the repositories of all cloned and forked repositories causing a large conflict as git is not able to align those histories.

Bachibouzouk:
> @Bachibouzouk Can you give a rationale for that? What I did is what the Driessen workflow proposes but instead of merging the release-branch back at release, I merged the master which is at the same commit. So those are equivalent.

I don't agree that they are equivalent. In the Driessen workflow, master is never merged back into develop. We should then maybe merge the release branch into develop.

I agree that to rewrite the history of master is not ideal.

MGlauer:
They are equivalent in the sense that they contain the exact same commits. So there is no semantic difference - just the name. But yes, I will just redo the merge with the release branch

PR #451, opened 2019-10-22
Feature/redesign on master

MGlauer:
This should fix the conflicts while retaining the changes in the original redesign


Bachibouzouk:
@MGlauer - I would fix the conflict in one single file (base.html) rather than perform a wild double merge master -> develop ->design-rework-welcome-page

PR #452, opened 2019-10-23
Merge Release/v0.5.0 back into develop

Bachibouzouk:
Solved the merging conflict choosing what was on HEAD` side


Bachibouzouk:
@MGlauer you can merge it if you approve it. I will not work anymore on the oep for the rest of the week and it would be good to merge it as soon as possible to avoid potential file conflicts with other PR :) Thanks

base/templates/base/base.html
```diff
@@ -105,18 +105,20 @@
...
+                        <li class="inactive"><a href="/faq/">FAQ</a></li>
+                        <li class="inactive"><a href="/contact/">Contact us</a></li>
+                        <li class="inactive"><a href="/legal/impressum/">Impressum</a></li>
+                        <li class="inactive"><a href="/legal/datasec/">Datenschutzerklärung</a></li>
+                        <li class="inactive"><a href="/legal/tou/">Terms of use</a></li>
+                        <li class="inactive"><a href="http://opendefinition.org/od/2.0/en/" target="_blank">Open Data {% bootstrap_icon "new-window" %}</a></li>
+                        <li class="inactive"><a href="https://okfn.org/" target="_blank">Open Knowledge  {% bootstrap_icon "new-window" %}</a></li>
+                    </ul>
+                  </div>
+            </nav>
```
MGlauer: I think this merge goes in the wrong direction. The deleted part is the one that is on master (and should be)

base/views.py
```diff
@@ -35,7 +35,7 @@ def get(self, request):
         return render(
             request,
             "base/index.html",
-            {"version": "%s.%s.%s" % (major, minor, patch), "changes": changes, "disableSidebar": True},
+            {"version": "%s.%s.%s" % (major, minor, patch), "changes": changes},
```
MGlauer: This parameter is also present in the new design on master but does not occur in `index.html`. Strange... maybe @johannwagner can say more.

base/templates/base/base.html
Bachibouzouk: Thanks for spotting that! I kept `develop` version when I should have chosen `HEAD` for the conflict resolution, I did a reset and force pushed the result of the new merge.

base/views.py
Bachibouzouk: Same comment as above

Bachibouzouk:
@MGlauer - I redid the merge and force pushed the result on this branch, now the changes should not be problematic anymore :)

PR #460, opened 2019-11-13
bugfix: nullable column in oedb in alembic

wingechr:
fixes problem with data insert via api (`null value in column "_message"`) by making column nullable

### DIFF
--- a/oedb_datamodels/versions/4ce486182650_nullable_message_really.py
+++ b/oedb_datamodels/versions/4ce486182650_nullable_message_really.py
@@ -1,0 +1,27 @@
+"""Make message nullable
+
+Revision ID: 4ce486182650
+Revises: b4e662a73272
+Create Date: 2019-04-30 09:04:34.330485
+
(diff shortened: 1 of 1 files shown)


Bachibouzouk:
@wingechr - it is finished ? Could you provide an example of commands which would generate the bug so that a reviewer can see if you fix works?

wingechr:
It's finished.

## Steps to reproduce
* primary database (oedb) is still operating at the version of the last release version (0.4.0)
* checkout of the development: `git checkout develop`
* upgrade the primary database with the new alembic system: `python manage.py alembic upgrade head`
* start the platform: `python manage.py runserver`
* use any client to insert data into a table: `POST <URL>/api/v0/schema/model_draft/tables/<TABLE>/rows/new ... `

This requests fails with the message: `IntegrityError('null value in column "_message" violates not-null constraint)`

This bugfix makes the alembic upgrade work the way it was intended to

Issue #466, opened 2019-11-14
Tabs and Spaces are mixed in securitysettings.py.default


christian-rli:
Indentation is inconsistent. Use spaces for everything.

Bachibouzouk:
Closed in #465

PR #467, opened 2019-11-14
Replace tab indentation with space indentation #466

christian-rli:
Makes securitysettings.py.default consistent (and prettier).

### DIFF
--- a/oeplatform/securitysettings.py.default
+++ b/oeplatform/securitysettings.py.default
@@ -16,11 +16,11 @@ token_test_user = '0'
 # https://docs.djangoproject.com/en/1.8/ref/settings/#databases
 DATABASES = {
     'default': {
-    'ENGINE': 'django.db.backends.postgresql_psycopg2',
-    'NAME': 'oep_django',
-    'USER': os.environ.get("OEP_DB_USER"),
-    'PASSWORD': os.environ.get("OEP_DB_PASSWORD"),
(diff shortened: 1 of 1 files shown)


oeplatform/securitysettings.py.default
```diff
@@ -16,12 +16,12 @@ token_test_user = '0'
...
     'default': {
-        'ENGINE': 'django.db.backends.postgresql_psycopg2',
-        'NAME': 'oep_django',
-	'USER': os.environ.get("OEP_DB_USER"),
-	'PASSWORD': os.environ.get("OEP_DB_PASSWORD"),
-	'HOST': 'localhost'                      
-	}
+    'ENGINE': 'django.db.backends.postgresql_psycopg2',
+    'NAME': 'oep_django',
+    'USER': os.environ.get("OEP_DB_USER"),
```
christian-rli: "OEP_DB_USER" could be changed to "oep_django_user" as is the placeholder in the README.md . Similarly "OEP_DB_PASSWORD" could be changed to "oep_django_password". Do you agree @Bachibouzouk ? This was mildly distracting during the tutorial, but nothing major.

christian-rli: Never mind. I see that this was addressed in the README right after.

Bachibouzouk:
Already addressed in #465

Issue #314, opened 2018-09-21
API for adding meta data


wingechr:
Can we also upload meta data via the api? Currently, it only seems to be possible on the platform via forms.

christian-rli:
API and dialect don't support uploading metadata yet, unfortunately. Implementation is high priority.

christian-rli:
"Implement API-support for metadata" - User Feedback
RequirementSpecificationID=17

wingechr:
first draft (without validation of metadata) implemented in 21b72eb93f5fa232d

Bachibouzouk:
@wingechr, @MGlauer if you are working on this, could you push the branch containing your work and link it to this issue?

MGlauer:
This was delayed due to #484 and #502. Now I can go back to testing this

MGlauer:
Closed in #477

PR #477, opened 2019-11-15
Feature/metadata api 314

MGlauer:
Adds metadata functionality to the API.


Bachibouzouk:
@MGlauer does this close the issue #314

wingechr:
Works, but two minor bugs in omi package (not oeplatform):

* allows writing of metadata without `resources` (omi parser does not fail). If you do that, the site crashes when trying to access the table in the data viewer
* when adding `resources` without `foreignKeys`, omi parser raises Exception

MGlauer:
This has been fixed in omi v0.0.4

PR #497, opened 2019-11-27
--do not merge-- Fix/comment and sort dataview view

Bachibouzouk:
Merge roadmap:

- Insert this work after #498 is merged

## Context 

In order to start working on the plot view I needed to understand how the views were structured, therefore I went into the code and :

- Add comments to all the function in the view.py file of the `dataedit` app. These describe what the view is supposed to show. 
- Rearrange the functions in the `view.py` module so that they are grouped more or less by category
- Remove templates which seemed unused

## Deleted html templates (they were not mentionned in any view)
- [x] dataedit_map.html
- [x] dataedit_revision_pending.html
- [x] dataedit_upload.html
- [x] list_field.html
- [x] url_list_field.html
- [x] map.html
- [x] tag_form.html
- [x] taggable_setting.html
- [x] tagged_field.html

### DIFF
--- a/dataedit/templates/dataedit/tag_editor.html
+++ b/dataedit/templates/dataedit/tag_editor.html
@@ -3,7 +3,7 @@
 
 {% block data_content %}
 
-{% if id != None %}
+{% if tag_id != None %}
     <h3>Edit Tag</h3>
 {% else %}
     <h3>Create Tag</h3>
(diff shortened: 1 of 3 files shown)


MGlauer:
As this already has the potential to cause major merge conflicts, we may use the opportunity, go the full mile and run black on this view.

Bachibouzouk:
> As this already has the potential to cause major merge conflicts, we may use the opportunity, go the full mile and run black on `view.py`

Very good idea :) !!!

wingechr:
reopen as needed

Issue #491, opened 2019-11-26
There is no graph library for bootstrap4 


MGlauer:
As described in #486 map and table functionalities have been replaced by more modern libraries. The old graph functionality is not yet implemented, though.

Bachibouzouk:
The new functionality should to to `backend.js`. It should retrieve the data via AJAX and create a plot in a div in dataedit/templates/dataedit/dataview.html.

MGlauer:
Below is how the old plot function looked like. The new one should follow a simiar scheme but usability improvements are always welcome. Maybe we should also integrate a filter functionality as in #498. 

![image](https://user-images.githubusercontent.com/7835111/69734443-45c69a00-112f-11ea-9647-a489245c547b.png)

Bachibouzouk:
I am trying to figure out where is the code which generates this view you show in the comment above @MGlauer. This was done in `dataedit/templates/dataedit/dataedit_overview.html` which you seem to delete and replace by `dataedit/templates/dataedit/dataview.html` in #498

MGlauer:
This was a recline view. Recline took a dataset and created multiple views. This functionality is not there anymore - as we are replacing recline. Therefore you will not be able to find it. The creation of the graphs was handled by recline internally.

MGlauer:
The data retrieval should not be too problematic. I can support you here if you want. From my perspective the real issue is that time series is heterogeneous. They can be stored horizontally or vertically (recline supported vertical data only). What we need is a form that allows users to select those series and trigger them to be plotted in the graph. I am not really sure how to design such a form, unfortunately.

Bachibouzouk:
It is not clear to me how I should retrieve the data using AJAX, are you suggesting to use the function `request_data` from `backend.js` @MGlauer ?

The code structure in `backend.js` should be discussed before I implement further functions (it could also have some comments for better reusability...). You seem to have an idea in mind, could you draw it schematically on a piece of paper (i.e. which functions could be reused for map, datatable and graph)?

MGlauer:
![_20191127_170007](https://user-images.githubusercontent.com/7835111/69739361-df924500-1137-11ea-9127-7f34c27afb9e.JPG)

Bachibouzouk:
Thanks for your schema @MGlauer :)

I think @Ludee wanted to be able to create several graphs resulting in a view like this :+1: 
[...]

Issue #486, opened 2019-11-20
Some major packages need to be updated


Ludee:
Identified packages that need an update:

- Bootstrap 3 -> 4
- Django [1.11 (LTS: Supported until at least April 2020) ](https://www.djangoproject.com/weblog/2017/apr/04/django-111-released/)
-> [2.2 (LTS: Supported until at least April 2022) ](https://docs.djangoproject.com/en/dev/releases/2.2.3/)
-> [3.0 (UNDER DEVELOPMENT)](https://docs.djangoproject.com/en/dev/releases/3.0/)
Under test in #485 
- recline.js -> leaflet & plotly ? (this issue is further described in #491 and solution is WIP in #499)
- datatables (this is WIP in #498)

Roadmap for merging : 

1. @MGlauer + @Bachibouzouk  : #499, then #498, then #485 
2. @4lm, @johannwagner : work on boostrap 4

MGlauer:
Tables use recline, too
- datatables

Bachibouzouk:
> Tables use recline, too
> 
>     * datatables

I am not sure what you mean? Is it a problem statement? or a todolist?

Bachibouzouk:
https://github.com/OpenEnergyPlatform/oeplatform/tree/feature/recline_alternatives

MGlauer:
> > Tables use recline, too
> > ```
> > * datatables
> > ```
> 
> I am not sure what you mean? Is it a problem statement? or a todolist?

That is a continuation of @Ludee's List

MGlauer:
Ok, I [replaced the tables](https://github.com/OpenEnergyPlatform/oeplatform/tree/feature/recline_alternatives) as @Bachibouzouk pointed out. Additionally, I also replaced the map-functionality. This whole process had some implications:

- Tables and maps are now displayed at the same time - Only the data displayed in the table is displayed in the data (thereby circumventing #369). 
- I integrated [jQuery Query Builder](https://querybuilder.js.org): This library was the only one delivering the functionality I was looking for. Unfortunately its transition to Bootstrap 4 is not done it. According to [this](https://github.com/mistic100/jQuery-QueryBuilder/issues/432), this can be worked around, tho. (Did that hopefully)
![image](https://user-images.githubusercontent.com/7835111/69442128-48e71200-0d4c-11ea-9b9f-8d87ef1dfccc.png)

I focused at functionality. The design of the functionality has do be done by those more capeable.

PR #498, opened 2019-11-27
Feature/recline alternatives

Bachibouzouk:
Merging roadmap : 

- Merge #499 first, then merge this PR


Bachibouzouk:
@MGlauer - what is still required before we can merge this work into `develop` ?

MGlauer:
> @MGlauer - what is still required before we can merge this work into `develop` ?

Essentially #491

Bachibouzouk:
> Essentially #491

For this I needed to understand how the views.py worked in `dataedit` and I added docstrings to the functions in #497. Can't we merge this PR and then I implement the graph in a subsequent PR?

MGlauer:
This would lead to us loosing functionality in the dev branch. I would like to avoid that. Also: The views are (for now) not part of this ticket. Tables and maps draw their data through the API and so should graphs.
The only thing, where a server-side handling would come into play, is when we add the view functionality. I would argue that we could leave that one out for now, if you add a form to select the data that should be plotted by the graph.

Bachibouzouk:
Ok. This is because in order to test if things works server side I wanted to generate a view with a plotly graph and see it

MGlauer:
I am not sure I understand. The only task done on the server side is collecting a part of the data that was requested via the API. The actual plotting is done in JS on the client side. So, I don't see what you want to plot on the server-side.

Bachibouzouk:
> I am not sure I understand. The only task done on the server side is collecting a part of the data that was requested via the API. The actual plotting is done in JS on the client side. So, I don't see what you want to plot on the server-side.

I see that I lack a precise vocabulary to express what want to do. Also, I probably did not understand what you meant by 
> Tables and maps draw their data through the API and so should graphs.

I would like to know what you would consider a solution to #491 (for my understanding it would involve changing something in the function `plot_view` in the `dataedit/static/dataedit/views.js`)

Bachibouzouk:
@MGlauer - if I want to test this branch, I don't have the same data table as the one you showed in [here](https://github.com/OpenEnergyPlatform/oeplatform/issues/486#issuecomment-557598007). Would it be possible to design a test dataset that one could use for test purposes?

MGlauer:
On this branch all the functionality has been moved to the `backend.js` as `view.js` mostly relied on recline and did not support the new libraries. Therefore, the new functionality should to to `backend.js`. It should retrieve the data via AJAX and create a plot in a div in `dataedit/templates/dataedit/dataview.html`. But let's move more detailed discussions to #491

Bachibouzouk:
This is what I see when I try the selection rules :
Screenshot from 2019-11-27 16-08-49

are they implemented yet?

MGlauer:
Interesting. It looks like the table is detached from it's header. I am not sure how this can happen but will look into this

MGlauer:
Also some of the filters do not work yet. (Because the API does not support them) I am working on it.

Bachibouzouk:
I build the table using https://github.com/OpenEnergyPlatform/examples/blob/master/api/OEP_API_tutorial_part1.ipynb

dataedit/static/dataedit/backend.js
```diff
@@ -227,6 +235,7 @@ function load_table (schema, table, csrftoken) {
         }
         if(table_info.geo_columns.length>0){
             map = L.map('map');
+            debugger;
```
Bachibouzouk: Belongs to the commit?

MGlauer: Oh, no...

PR #499, opened 2019-11-27
Add plotly graph

### LINKED ISSUE #491: There is no graph library for bootstrap4

Bachibouzouk:
Closes #491 

The model `View` will contain the api request arguments for a given view on the data. This way one can store only how to make the api call in the django db. This could be useful if we want to allow several graphs views to be saved, saving only the search criteria will not take too much space on our server.

## Still remaining to do

- [ ] get the list of columns of a table from the api


dataedit/views.py
```diff
@@ -758,6 +759,26 @@ def view_delete(request, schema, table):
...
+def create_graph(request, schema, table):
+
+    if request.method == 'POST':
+        print(request.POST, schema, table)
+        # TODO save an instance of tablegraph here
+        return redirect(
+            "/dataedit/view/{schema}/{table}".format(schema=schema, table=table)
+        )
+    else:
+        # TODO get the columns from the schema and the table
```
Bachibouzouk: @MGlauer , @gnn do you know a function I could use for that?

Bachibouzouk:
@MGlauer - I would need help to get the data from the DB in ajax reusing the function you already wrote for the datatable in #498 :)

dataedit/models.py
```diff
@@ -62,6 +62,16 @@ class Meta:
...
 
 
+class TableGraph(models.Model):
+    name = CharField(max_length=50, null=False)
+    table = CharField(max_length=1000, null=False)
+    schema = CharField(max_length=1000, null=False)
+    column_x = CharField(max_length=200, null=False)
+    column_y = CharField(max_length=200, null=False)
+    filters_x = CharField(max_length=1000, null=False)
+    filters_y = CharField(max_length=1000, null=False)
```
MGlauer: Why are filters for x and y separated? What does it mean to have a filter on x but not on y? Why are they strings?

Bachibouzouk: I just added the filter as an exemple, I don't actually know what the filter field should contain

dataedit/views.py
MGlauer: Is it necessary to do this server-side? The table functions query all columns via the API and [process their name and datatype](https://github.com/OpenEnergyPlatform/oeplatform/blob/3440f5e5b89a31d366fd2e99115bc92c44d4f930/dataedit/static/dataedit/backend.js#L231). Could you use this instead?

Bachibouzouk: What I need to have is the list of columns of the table so that the dropdown for x and y are populated. If I want to use forms and not write js code in the template I need to have this list here, but the list could be provided via the GET.

I will describe my idea for the model of TableGraph in the PR's description

Bachibouzouk: @MGlauer, I described it

dataedit/views.py
```diff
@@ -758,6 +759,37 @@ def view_delete(request, schema, table):
...
+            is_default=request.POST.get('is_default')
+        )
+        gview.save()
+
+        return redirect(
+            "/dataedit/view/{schema}/{table}".format(schema=schema, table=table)
+        )
+    else:
+        # TODO get the columns from the schema and the table
+        columns = [
```
Bachibouzouk: @MGlauer, @gnn - here how could I get the columns of the table with the table and schema's name. Is there a function of the api for that?

MGlauer: You could use describe_columns from actions.py like this:
describe_columns(schema, table).keys()

Bachibouzouk: Thanks :)

Bachibouzouk:
@MGlauer can you test it? I can create graph views but then the graph is not displayed.

I would propose to have at least one view of each type (default) and then be able to update the possible view by clicking on buttons. (@Ludee is this what you had in mind?)

------------------------------------------------------------------
section table
[dtview1-default] [dtview2] [dtview3]
- datatableview1
------------------------------------------------------------------
section map
[mapview1] [mapview2-default] [mapview3]
- mapview-2
------------------------------------------------------------------
section graph
[gview1-default] [gview2]
- gview1
------------------------------------------------------------------

MGlauer:
I am not sure what the default for a graph could be. How do we determine which columns to plot? For maps we could check the column types and names and look for GIS-types and "lat"/"lon" in the names.

There is already a 'default' view for tables (see in views object passed to the dataview).

Bachibouzouk:
> I am not sure what the default for a graph could be. How do we determine which columns to plot? For maps we could check the column types and names and look for GIS-types and "lat"/"lon" in the names.
> 

I thought that the default when there is no already-defined default view is no view at all, just the option to create one (in the example above it was implicit that a user had already created a view, which would automatically become the default view; it was not clear I concede)

So for a brand new upload the first view would be like :

------------------------------------------------------------------
section table
- datatableview-default
------------------------------------------------------------------
section map
[add new map view]
------------------------------------------------------------------
section graph
[add new graph view]
------------------------------------------------------------------

MGlauer:
I think the map view and graph view should be mutually exclusive. I do not see a good use case to have both of them in the same view. Therefore I would propose to merge section map and section graph.

MGlauer:
But this is a question of design not functionality and therefore not really part of this PR.

PR #505, opened 2019-12-05
Rename CONTRIBUTE to CONTRIBUTING

### LINKED ISSUE #504: CONTRIBUTE should be renamed CONTRIBUTING
Github has some magic for us (notification to new users etc...) when it sees the special name CONTRIBUTING

Bachibouzouk:
Closes #504


Bachibouzouk:
@4lm - if you can review fast it would be good as it is only a small change :)

PR #509, opened 2019-12-09
Update README for windows env variable

### LINKED ISSUE #508: The windows install instructions failed with a user
## Problem

the environment variables defined in command line were only updated after reboot of the system

## Proposed solution

write in the README that it is advisable to set the environment variable via menus (could use this [link](https://www.computerhope.com/issues/ch000549.htm))

Bachibouzouk:
Closes #508

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,12 +77,12 @@ Create the environment variables `OEP_DJANGO_USER` and `OEP_DJANGO_PW` with valu
 For default settings, you can type the following commands
 
 - On windows
-
-     before you can set environment variables in your terminal you should first type `cmd/v`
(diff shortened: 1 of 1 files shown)


Bachibouzouk:
> I would omit the "cmd/v" part since it did not work as expected. The best way would be to set the env-variables via the dialog-window and then restart the machine.

One must restart?

Bachibouzouk:
> I would omit the "cmd/v" part since it did not work as expected. The best way would be to set the env-variables via the dialog-window and then restart the machine.

It did not work as expected on one windows machine, I'd rather keep it nevertheless to cover a larger spectrum of solutions for windows cases

jh-RLI:
Yes, one must restart, there are other ways but they are not always reproducible.

PR #510, opened 2019-12-10
Connect plotly graph to API

MGlauer:


MGlauer:
I can. But for this I need to know how the column names are passed to the script.

Bachibouzouk:
> I can. But for this I need to know how the column names are passed to the script.

I was thinking to save the column names in the options field of the View instance under `x` and `y` (you can see it in #499, I pushed my latest changes)

MGlauer:
Then I will include it in this pr.

MGlauer:
Ok, the columns for graphs are loaded from the `current_view` variable now.

PR #489, opened 2019-11-22
Feature/tutorials

Bachibouzouk:
Closes #478 

- [x] Create new tutorial using markdown
- [x] Add new tutorials to list and detail view 
- [x] Edit tutorial
- [x] Delete tutorial
- [x] Markdown is converted to HTML
- [x] Add markdown editor
- [x] exclude public user from C ~~R~~ UD funktions
- [ ] View for image upload
- [x] Documentation
- [ ] Frontend like the mockup (see [this)](https://github.com/OpenEnergyPlatform/oeplatform/issues/462#issuecomment-553813460)

PR #516, opened 2020-01-08
Handle video content in Tutorials section

jh-RLI:
#489 must be merged first!

- Add model field and form field as an input field for URL
- retrieve video source as URL from django database
- Embed video in HTML 

run migration after merge!

Note: I did not use rebase :/ first relevant commit ist 52318bf


jh-RLI:
This branch is fixed in PR #615.

PR #534, opened 2020-03-05
Reworking Factsheets View

johannwagner:
I reworked the view for the factsheets for Models and Frameworks.
I moved the filter methods within the sidebar to prioritise the data view by reusing some elements from the data view.

Two Concerns: 

- Not sure, what to do with the column filters. They are less bulky now, but I did not find a good element within bootstrap to implement them.
- I could not test scenarios, because I was not able to create one.


MGlauer:
Is this ready for review?

johannwagner:
@MGlauer It is.

PR #540, opened 2020-03-10
Refactored Map and Graph View creation to use bootstrap forms

johannwagner:


johannwagner:
This is garbage. I did not use django-bootstrap.
I will redo this.

PR #581, opened 2020-03-19
Migrated View Forms to Bootstrap 4

johannwagner:


MGlauer:
I think the map-form should be split into two different forms. One possibility is that users select at the beginning whether lat/lon or geom will be used and then the respective form is shown.

PR #582, opened 2020-03-19
Add Textarea insted of Inputfield

jh-RLI:
When editing a "broken" JSON-string the metadata string is now shown in a textarea instead of a simple input field.


jh-RLI:
@MGlauer I have tested the fix and no errors have been noticed. Generally, I find the implementation a bit workaround-like. As it seems django is not used very much.

PR #599, opened 2020-04-02
Feauture/ffs implement feedback

jh-RLI:
Implementing Framework-Factsheet developer feedback


jh-RLI:
I have implemented the feedback from a review for framework factsheets here. I came across the following problem:

I cannot find the source of the tooltips for the fields: auto_model_generator, variable_time_step and variable_rolling_horizon. In modeview/models.py there is a help_text for the variables and the desired content is available there. However, the tooltips appear to be overwritten. 
e.g. the tooltip for auto_model_generator is help_text="Is an auto-model generator available that transfers an input file into a model?"but in the frontend this tooltip is shown:
image

Does anyone have an idea? @MGlauer @johannwagner

jh-RLI:
@MGlauer It would be great if we can quickly merge this PR into the production environment. @Ludee

jh-RLI:
@Ludee I have figured out why the text (as described in my [comment)](https://github.com/OpenEnergyPlatform/oeplatform/pull/599#issuecomment-610980791) is changed. See [this].(https://github.com/OpenEnergyPlatform/oeplatform/blob/ee9b161db0b78c3fe3b6c00caff42772926051ec/modelview/forms.py#L30)
To fix the unwanted behavior without affecting the functionality for other help_texts, one solution is to add another parameter. With the parameter, the forms-view can decide if the helptext should be changed or not.

MGlauer:
It seems that the entries for `es_other` are not displayed in the framework factsheet.

jh-RLI:
@MGlauer fixed this and noticed that it shows "other" instead of the value given in the column!

MGlauer:
Yes, this is because the default behaviour of the checklist tag is that it shows the `verbose_name` of the corresponding field. If you want to show something different, you have to specify it. Instead of `es_other` you should pass `es_other=es_other`. [See here](https://github.com/OpenEnergyPlatform/oeplatform/blob/2296056748b81c047c613c3f37cde07139465d61/modelview/templates/modelview/model.html#L262)

MGlauer:
As an explanation: Normally, it is just checked, whether the value in the given field is true or not. If it is, the `verbose_name` is displayed. If you pass `A=B` instead, `B` is displayed iff `A` evaluates to `True`.

jh-RLI:
@MGlauer please re-review this

PR #615, opened 2020-08-04
Inital implementation of video content within tutorial section

johannwagner:
This is another pull request to implement videos within the tutorial section.
I cherry-picked the other commits from the broken branch and squashed them together.

Right now, it is not fully discussed, if this should be merged within this state.

I propose to close https://github.com/OpenEnergyPlatform/oeplatform/pull/516 in favor of this PR.


christian-rli:
This branch works on my local instance and I got a test video to play. Is there a particular reason why it's still a work in progress? I'd like it to be part of the next release :)

johannwagner:
No, there is no reason, why this is not able to merge. I think, this is pending on @MGlauer.

MGlauer:
Is there any documentation on how to use the form to include video content? At least in the "Add tutorial" form it does not seem obvious.

MGlauer:
Ah, ok - when linking to a youtube video you would need the embed link. I guess this is not something any unexperienced user would do. Is there any way to make it more convenient?

johannwagner:
Yup, I did a small google research and you may be able to convert it, if it is not an embed link. Should we try this? 

Reference: https://stackoverflow.com/questions/29781974/convert-youtube-link-into-an-embed-link

MGlauer:
Yes, this would improve the usability massively.

johannwagner:
I will take care of this feature request.

christian-rli:
thanks, you two.

MGlauer:
I guess this can be easily done with a simple regex: 

```
^https:\/\/www\.youtube\.com\/watch\?v=(?P<id>[A-z0-9]+)$
```
and then
```
"https://www.youtube.com/embed/{id}"
```

Question is, whether there are any services that are a bit less thirsty when it comes to personal data. @christian-rli any ideas?

johannwagner:
We decided to implement YouTube as a first step and may add other video on-demand streaming portals later on.

PR #628, opened 2020-09-07
Creating a new app  "Metatool" in django oem_creator

TuPhanRLI:
Creating an interface for setting  a metadata syntax based on  oemetadata/metadata/latest/metadata_key_description.md and the template in oemetadata/metadata/v140/example.json
unnecessary paths in static


jh-RLI:
@TuPhanRLI I think we need to include the [omi parser](https://github.com/OpenEnergyPlatform/omi/blob/06f840907e39b79348788f7bfcdccb13a75e79e7/src/omi/dialects/oep/parser.py#L186) to check if the resulting JSON string is valid with the oep. 
The Download JSON button is not working for me.

PR #630, opened 2020-09-14
feature/tutorial-titles

johannwagner:
@MGlauer asks me to parse some human-readable titles from the html, which is generated from the Notebooks.
Therefore we introduced a `meta.json` file, which is generated at **build** time, which includes a list of files and additionally parsed titles and some meta information, which was hardcoded beforehand.

Therefore, we can now add new static tutorials without touching code.
Also, we now have human-readable titles.

### DIFF
--- a/base/management/commands/notebooks.py
+++ b/base/management/commands/notebooks.py
@@ -23,7 +23,7 @@ class Command(BaseCommand):
 
         if options["command"] == "build":
             # We need to generate the HTML files from the python notebooks.
-            os.system('jupyter nbconvert examples/**/*.ipynb --output-dir=examples/build --to html --template examples/templates/openenergyplatform.tpl')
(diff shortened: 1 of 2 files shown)


TuPhanRLI:
@johannwagner I try to open the tutorial direction of ur branch , but tittles didn't really show up at the tutrial section.

johannwagner:
You need to rebuild the tutorials, because we generate some meta information in this compile step. There is some documentation in the Readme.md about this.

johannwagner:
In this PR we introduce some manual steps. Is this a problem within a server deployment? @MGlauer

han-f:
Would this issue resolve the display of tutorial titles here: 

http://toep.iks.cs.ovgu.de/tutorials/ ?
If yes, it would be great if this could be fixed for the next release.

PR #632, opened 2020-09-22
Feature/csv upload 346

wingechr:
upload wizard first working draft (UI needs some poslishing)


MGlauer (commented):
The wizard seems to work fine. I tried to break it, but couldn't ;P But I found some unexpected behaviours: 

1. If nullable is checked and I select "Primary key", nullable is unchecked automatically, but I can still re-check it
2. Some invalid table names trigger a "Not found" error instead of a proper error message pointing to the table name. But I guess this is an API-bug

wingechr:
> 
> 
> The wizard seems to work fine. I tried to break it, but couldn't ;P But I found some unexpected behaviours:
> 
>     1. If nullable is checked and I select "Primary key", nullable is unchecked automatically, but I can still re-check it

I added functionality to disable nullable checkbox if PK is selected

>     2. Some invalid table names trigger a "Not found" error instead of a proper error message pointing to the table name. But I guess this is an API-bug

I added simple (visible) validator for idenfier names, but i won't stop users to try and use the api with invalid input. especially with checking sql datatypes, there is too many thinsg to consider.

MGlauer (approved):

Issue #167, opened 2016-10-14
Data view: Can / should the unit be part of the data view?


EvaWie:
We have to avoid the impression that the unit is part of the column name.

christian-rli:
The specified unit is visible in the metadata. It could be displayed in a mouseover though.

wingechr:
resolved with https://github.com/OpenEnergyPlatform/oeplatform/pull/653

Issue #134, opened 2016-09-11
Data view: design suggestion to increase the understanding


EvaWie:
- order of the elements on the page: Name of Table, Tags, explanation/help (hide symbols +/-), "Show also unconfirmed data" (0), Data,  "Data is displayed using recline.js", comments (1), "Commit your changes" box (2), meta data, Revision (3)

0) Rename and add explanation. See also #147

1) Make the text ">>Commments" more clear: "Show the table explaining the comments on the rows". Can we open it in another tab?

2) put "upload csv" into the grey "commit your changes" box under the explanatory text (+ "You can make change in the table above and / or upload additional rows from a csv file.") See also #147.

3) Make the text more clear: "You can download THE CURRENT OR AN older revisioN of the data stored On the platform. Each revision has a unique id which can be used for citations. In order to save memory ON THE SERVER THE RELEASES ARE NOT DIRECTLY are available. You can request a dump of the current table OR A FORMER VERSION by..."
What about the check box to order related tables, too?

A) Bug: displaying: source on top of the meta data instead of in the table

B) Lastest revision mark it as "lastest / current revision/ release"  See also #143.

C) Will untested data be displayed in another font colour? See also #147.

christian-rli:
"Redesign the data view: Display the column descriptions from metadata." - User Feedback and part of the requirement specification. 

https://github.com/OpenEnergyPlatform/oeplatform/issues/226 and https://github.com/OpenEnergyPlatform/oeplatform/issues/167 are connected to this issue.

RequirementSpecificationID=49

wingechr:
finished with #167 and #226

PR #653, opened 2020-10-27
Feature/dataview popups 167

wingechr:
Add popover in data viewer for column description and unit (#134, #167)


dataedit/static/dataedit/backend.js
```diff
@@ -262,21 +262,44 @@ function load_view(schema, table, csrftoken, current_view) {
...
+            url: '/api/v0/schema/' + schema + '/tables/' + table + '/meta'
         })
-    ).done(function (column_response, count_response) {
-
-
+    ).done(function (column_response, count_response, meta) {
+        var columnDescription = {}; // name -> description
+        try {
+            meta[0]['resources'][0]['schema']['fields'].map(function(fld){
+                columnDescription[fld.name] = fld.description || "";
```
MGlauer, line 282: As far as I understand this would fail if a single field does not have a "name" property. This may happen due to a faulty metadata string and should be caught.

MGlauer (changes requested):

dataedit/static/dataedit/backend.js
wingechr: no, that would not throw an error, it would only set columnDescription[undefined]

wingechr: i guess it would only break if fld is not an object

MGlauer (approved):

PR #657, opened 2020-10-28
fix/tutorial_detail-view escape html

### LINKED ISSUE #656: Tutorials from markdown look different in preview then in final version
Steps to reproduce:

* New tutorial:

```markdown
<a name="anchor">i am an anchor</a>
[i am a link to the anchor](#anchor)
```

* Preview (is ok)

preview

* Submit (no raw html):

rendered

### LINKED ISSUE #674: Tutorials app: activate safe_mode
Html is not escaped as expected when creating a new tutorial using the markdown editor -> save tutorial.

jh-RLI:
Fix html escaping. Experience new bugs: 

- html markdown code block is escaped
- checklist not rendered as expected

Closes #674


wingechr (commented):
this does not yield the intended results. raw html still cannot be used. The Only way to achieve that would be `safe_mode=False`, but that might be a bad idea for security reasons?

jh-RLI:
I don't think we will support raw html, as this will trigger the risk of XSS. Most editors that allow raw HTML only allow certain elements. Unfortunately I don't have time to take a closer look ATM.

wingechr:
i suggest to delete the branch and close the issue as`will not be fixed`

jh-RLI:
I think we need the html escape anyway, so far it was defined wrong and therefore it was not active. 

old:
`markdowner = Markdown(safe_mode="escape")`
->this does not activate safe mode

new:
```
  markdowner = Markdown(safe_mode=True)
  markdowner.html_removed_text = ""
```
So I would merge the branch but close the issue as wont-fix. I the link to the issue.

PR #663, opened 2020-10-30
Feature/new meta editor

wingechr:


MGlauer:
@wingechr Can I review this now? It is still marked as WIP.

wingechr:
@MGlauer yes

dataedit/templates/dataedit/dataview.html
```diff
@@ -117,14 +116,21 @@ <h4>API Usage</h4>
...
+  <hr>
+
   <h4>
     Meta Information
   </h4>
-  <div class="btn-group btn-group-sm">
-    <button class="btn btn-success" onclick="downloadMetadata();">Download JSON</button>
+  <div class="mb-2">
+    <button class="btn btn-sm btn-success" onclick="downloadMetadata();">Download JSON</button>
+    <a href="{{table}}/meta_edit" type="button" class="btn btn-success btn-sm ">Edit</a>
```
MGlauer, line 131:
```suggestion
    <a href="{{table}}/meta_edit" type="button"  class="btn btn-success btn-sm {% if not can_add %} disabled {% endif %}">Edit</a>
```

dataedit/templates/dataedit/dataview.html
```diff
@@ -117,14 +116,21 @@ <h4>API Usage</h4>
...
+  </div>
+  <hr>
+
   <h4>
     Meta Information
   </h4>
-  <div class="btn-group btn-group-sm">
-    <button class="btn btn-success" onclick="downloadMetadata();">Download JSON</button>
+  <div class="mb-2">
+    <button class="btn btn-sm btn-success" onclick="downloadMetadata();">Download JSON</button>
```
MGlauer, line 130:
```suggestion
    <button class="btn btn-sm btn-success" onclick="downloadMetadata();" data-toggle="tooltip" title="Download metadata in JSON format">Download JSON</button>
```

dataedit/templates/dataedit/dataview.html
```diff
@@ -117,14 +116,21 @@ <h4>API Usage</h4>
...
+    <a href="{{ table }}/permissions" type="button" class="btn btn-danger">Permissions</a>
+  </div>
+  <hr>
+
   <h4>
     Meta Information
   </h4>
-  <div class="btn-group btn-group-sm">
-    <button class="btn btn-success" onclick="downloadMetadata();">Download JSON</button>
+  <div class="mb-2">
```
MGlauer, line 129:
```suggestion
  <div class="mb-2" {% if not can_add %} data-toggle="tooltip" title="You need write permissions on this table to edit metadata" {% endif %}>
```

MGlauer (changes requested):
The edit button works even if the user can not edit the metadata. We should disable it in this case and show a tooltip explaining the situation. But tooltip don't work on disabled inputs - so we have to add it to the top level element and add a tooltip to the other button to prevent the same tooltip showing on both buttons.

dataedit/templates/dataedit/dataview.html
MGlauer: We can add the `not can_edit` conditional here, too. But I feel that a tooltip could improve this button in general.

MGlauer (approved):

PR #716, opened 2021-01-06
reintroduced the datapackage download button, and remove duplicated permissions button

### LINKED ISSUE #711: Datapackage download link has disappeared
Previously we were able to download the csv of a table and a data package.
Now the possibility to download the data package has disappeared. We would like to demonstrate this functionality on January 20th, so it would be very good to get it back. 

Not sure who is the best to assign - please adjust accordingly.

jh-RLI:
The "download datapackage" button was lost or confused with the permissions button. I removed the duplicate permissions button and reintroduced the download.

Closes #711


jh-RLI:
@MGlauer this seems to be implemented already :) I think this PR can be closed.

PR #727, opened 2021-01-15
Hotfix/upload wizard column mapping

wingechr:
hotfix for final workshop: fix column mapping and minor tweaks in upload wizard (client side javascript only)


MGlauer:
Has been merged into hotfix branch

Issue #717, opened 2021-01-06
Model factsheet - Minor Bugs


Ludee:
I am testing and updating the factsheets and I found some minor bugs and problems in the model fact sheet that needs to be fixed:

- **Software** | External optimizer

When checkbox is _yes_ a text field is displayed. The text is not displayed in the factsheet after submission.
Bug is in _add=model.external_optimizer_yes_text_?
-> Allow multiple entries in form

- **Basic Information** | Link to User Documentation

The URL is broken with an internal URL and {}.
see: https://openenergy-platform.org/factsheets/models/32/
Bug in _code_documentation_ or url check function in _model_snipped.html_ ?

- **Basic Information** vs **General Information**

It's _basic_ in the factsheet but _general_ in the edit.
Please change to **General Information**

- Order of cards

Factsheet: ~~Basic~~ General; Openness; Software; **References**; Coverage; Math; Integration
Edit: General Information; Openness; Software; Coverage; Math; Integration; **References**; Tags

Please harmonise to:
General Information; Openness; Software; Coverage; Math; Integration; References; Tags
-> Move references in models.html to the end

- Expand first card General Information when factsheet is opened

- Cards in edit mode

There are no gaps/ spaces between the cards in edit mode.
Please add spaces to improve readability.

jh-RLI:
Added padding and border to the section headings:
![image](https://user-images.githubusercontent.com/38939526/103946879-8452c480-5137-11eb-8d98-be22d43da9df.png)

Do you think the frame is good or should I just add spacing in between (padding)?

jh-RLI:
See the PR above^

jh-RLI:
Fixed all except the Software | External optimizer and Basic Information | Link to User Documentation. Will work on it in the next few days.

han-f:
Just came across this and think padding would be nice. Also the border seems black which quite jumps at you - have it more in the dark blue theme instead?

jh-RLI:
I have removed the border again, because this does not look good on the different screen sizes (responsive). But I have added a little more padding so that the distance between the words is larger.

jh-RLI:
I have noticed that some fields are not displayed in the framework factsheets view. The fields are only visible when you are in edit mode. I have added the fields again. 

Fields added to openness:
- user_documentation
- installation_guide
- link_to_code_documentation
- skills_basic
- skills_advanced

jh-RLI:
[...]

PR #760, opened 2021-06-15
Release/0.9.1

christian-rli:
Joint release in first OEP developer meeting.

### Bugs
- Fix [#717: https://github.com/OpenEnergyPlatform/oeplatform/issues/717]
- Fix bug where entries in framework factsheets can be edited but are then not displayed.
- Fix tests for CI

### Features
- api command for users to move tables between schemas


christian-rli:
@MGlauer : The release branch now contains the updated requirements and is ready for testing on toep. The latest commits don't show up in this PR. However they are in the release branch (https://github.com/OpenEnergyPlatform/oeplatform/tree/release/0.9.1) and here that that's enough. I also verified that info looking into another issue (https://github.com/github/hub/issues/198 "you don't need hub to update pull requests. Pull requests are tied to the remote branch they're opened from")

Please let @wingechr know when it's ready for testing and @wingechr please let me or both of us know when I can commence with the release :)

wingechr:
looks ok

wingechr (approved):
looks ok

christian-rli:
> looks ok

Nice. Sorry for probing, but did you already test it on toep? I would just be impressed by the speed.

wingechr:
> 
> 
> > looks ok
> 
> Nice. Sorry for probing, but did you already test it on toep? I would just be impressed by the speed.

only on my local system

MGlauer:
First Feedback: The current version does not work with Django 3.0.* but it does with version 3.2.*. This should be amended in the requirements.txt

MGlauer:
The test server has been updated

wingechr:
> 
> 
> First Feedback: The current version does not work with Django 3.0.* but it does with version 3.2.*. This should be amended in the requirements.txt

I think we should choose one of two options:

* A) we fix all versions in requirements so that the environment is predictable
* B) we always upgrade to the currently latest version allowed by requirements at the time of the test / release

I assumend B) was already the case, but apparently not? This would also be my preferred option. Any thoughts?

christian-rli:
I also prefer option B and thought that this was the case. So I'd suggest to go with that from now on.

wingechr (approved):

MGlauer:
There are arguments to make for both options.
Option B guarantees that important security updates are installed as soon as they are available, *but* having the cutting edge version of any software package may introduce incompatibilites and even new security issues.

That is why most production environments settle with option A, in which you have a fixed set of packages that are thoroughly tested and work well together. Yet, there has to be some entity that keeps track of important security updates and manages necessary upgrades.

MGlauer:
But this is probably something that should be discussed in a different place. Still, here is some read how different linux distributions handle this issue: https://averagelinuxuser.com/rolling-vs-fixed-linux-release/

christian-rli:
> There are arguments to make for both options.
> Option B guarantees that important security updates are installed as soon as they are available, _but_ having the cutting edge version of any software package may introduce incompatibilites and even new security issues.
> 
> That is why most production environments settle with option A, in which you have a fixed set of packages that are thoroughly tested and work well together. Yet, there has to be some entity that keeps track of important security updates and manages necessary upgrades.

Thanks for the comments/explanation! I put it on the agenda for the next OEP DEV meeting.

christian-rli:
I also tested moving tables now and the release can go on.

christian-rli:
I checked the release description file, but final steps aren't fully described. What comes first - inform @MGlauer / merge into master / push release button? i'd be happy about a 3 line description or a quick call any time today.

MGlauer:
Production has been updated

Issue #763, opened 2021-06-17
Serving OEO classes: Readable by humans and machines


l-emele:
Via the OEO DEV mailing list we got the following feedback for the OEO:
> Unfortunately the OEO is still not usable (for Linked Data). Just have a look at http://openenergy-platform.org/ontology/oeo/OEO_00230002 - nothing there. I hope you now understand that it is crucial, to return sth. machine and human readable for every OEO id in order to be recognized and accepted as an ontology and you now also understand the direct practical reasons for it when it comes to actually using the ontology when using this "human-incomprehensible" IDs.

So it basically this asks that the class IRIs resolve to some kind of webpages for the specific pages and not just serving the content of http://openenergy-platform.org/ontology/oeo/

We discussed that today in the OEO DEV meeting. This is not a problem of the ontology itself, but if we want to have the functionality of serving the classes it would be a topic for the OEP. This should be discussed at the OEP DEV and/or the LOD_GEOSS meetings.

l-emele:
@christian-rli : I assigned this issue to you are both the bridge between the OEO and OEP and the bridge between SIROP and LOD_GEOSS.

christian-rli:
@Adel-Mem this is the issue we talked about in the OEP-DEV meeting. See how far you can get and we'll talk about it in the next meeting. Thank you for volunteering :)

adelmemariani:
I added the functionality to display sub/super classes for a given class. I think we should add the names for the classes. Also if it is needed, I can work on a new user interface to show the classes in a tree or graph view.

l-emele:
A tree or graph view would be imho great!

han-f:
I agree with @l-emele, a tree or graph view would be a great value added!

han-f:
Is this the correct issue to post the ideas we had in yesterdays (28/07/21) SIROP Jour Fixe when @adelmemariani showed us the progress that he has made on displaying OEOs terms on the OEP  (great value added!)? 
Here are the ideas from yesterday: 

- For each term its definition should be displayed
- Further items that should be displayed best to be discussed in an upcoming oeo-dev meeting.

adelmemariani:
[...]

PR #773, opened 2021-07-20
Display OEO sub/super classes #763

adelmemariani:
Since the changes have been done on a macOS machine, it is recommended to remove the '.DS_Store' files that are created by macOS.


christian-rli:
This didn't break anything for me and @adelmemariani mentioned to take this as a backup when the newer updates aren't ready yet. Also, we need this to help fixing the history of other branches.

PR #779, opened 2021-08-09
Feature/oep review

adelmemariani:
We need to run: 'python manage.py makemigrations' and 'python manage.py migrate' before running the server.


johannwagner:
This is something you generate on your machine. This will create migration files, which also should be added to GitHub.

christian-rli:
I am looking through PRs and try to resolve them before the next release.  What's missing in order for this to be resolved. Just running migrations and committing them?

PR #787, opened 2021-08-23
Add meta-schemas to schema creation

### LINKED ISSUE #786: Meta-schemas are not created properly
It seems that a fresh installation of the most recent version does not create the meta-schemas anymore.

MGlauer:
Closes #786


christian-rli (approved):
Schemas are created for me locally (albeit not shown directly, because there are no tables within). Not sure this should be different, but broke and I believe in your code @MGlauer

PR #790, opened 2021-09-01
OEP viewer

adelmemariani:
I removed client-side routing within the React App since they are not compatible with the Django routing scheme. I also added to source code for the front-end. I expect the build version of the modified code should work.


christian-rli:
Do I need to do something special to get this running? When clicking on OEO in the ontology menu I get a 404 error. 
<h1>Page not found <span>(404)</span></h1>
    
    

Request Method: | GET
-- | --
http://127.0.0.1:8000/ontology/oeo/
ontology.views.OntologyOverview

I do have a local ontoloy folder with a version of the oeo that @adelmemariani sent me before.

adelmemariani:
> http://127.0.0.1:8000/ontology/oeo/

I think with the latest update, you can see the graph/tree viewer with this URL locally: http://localhost:8000/oeo_viewer/
It worked in my local machine. When clicking the OEO, you should see a page similar to the following image:
Screenshot 2021-09-01 at 19 35 35

christian-rli (approved):
Looks good and works locally for me now :)

PR #795, opened 2021-09-10
Feature/oeo viewer

adelmemariani:
Fix the URL for the viewer. Add link for the viewer in OEO the page.


christian-rli (approved):
Thank you @adelmemariani . This works and the link is good for now :)

PR #810, opened 2021-09-20
Add migrations for dataedit, modelview and tutorials

adelmemariani:
Migration files, Feature/oep review #779


christian-rli (approved):
Migrations run thourgh clean. Website and database work afterwards :)

PR #822, opened 2021-10-27
Feature/factsheets bugfix

adelmemariani:
Before deleting an item from factsheets, it asks for the user's confirmation.


christian-rli:
which issues does this PR address @adelmemariani ? It's good practice to both mention it in the PR description and also to reference issues at the end of our commit messages by use of hashtag and the issue number.

PR #877, opened 2022-01-07
remove prints

adelmemariani:
Clickable links in model calculations (inputs) in factsheets alpha. The links refer to the corresponding table.


adelmemariani:
Since in the .gitignore file we have an entry as 'static/', the CSS file that I added for the styles in the factsheet alpha, is not pushed to the repository. Therefore, I think I should upload it manually to the server. Any opinion on how? :)

Issue #746, opened 2021-05-11
Explanatory text under 'Schemas' is missing


han-f:
I received feedback from a potential user of the OEP that it would be helpful to have a short text under 'Schemas' for explanation as otherwise this title is not self-explanatory for "Quantitists" and "Qualitists".  
![grafik](https://user-images.githubusercontent.com/36844029/117811058-6c638c80-b260-11eb-9d00-c50e91bd8f9e.png)

Suggestion: _Schemas organise the available data according to specific topics. The topics are shown and described in the table below._

jh-RLI:
I would vote to leave out "Schemas" and use something like "Data Catalog", but adding a description as you suggested is definitely good to have.

han-f:
> I would vote to leave out "Schemas" and use something like "Data Catalog", but adding a description as you suggested is definitely good to have.

Yes that would even be better, I somehow thought the heading was fix :-)

han-f:
New suggestion:
Heading: _Data Catalogue_
Explanation: _Data on the Open Energy Platform are organised according to specific topics. The topics are shown and described in the table below._

jh-RLI:
Schemas got renamed to topics that could be self-explanatory? If not, open a new one.

PR #940, opened 2022-03-30
Rework presentation of schemas 

jh-RLI:
closing #746

Convert data presentation method in "Database" from html table to bootstrap cards using flexbox layout.

Result (using mock data):
image

### DIFF
--- a/dataedit/templates/dataedit/dataedit_schemalist.html
+++ b/dataedit/templates/dataedit/dataedit_schemalist.html
@@ -2,28 +2,42 @@
 {% block title %} - Schemas{% endblock %}
 
 {% block site-header %}
-    <h2 class="site-header">Schemas</h2>
+<h2 class="site-header">Data categories</h2>
 {% endblock site-header %}
 
 {% block data_content %}
 <h3>Topics</h3>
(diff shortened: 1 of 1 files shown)


bmlancien:
@jh-RLI I started to work on a tabular view, but actually, a card view might be interesting too, since there are not too many items (I can see 13 right now). Once I'm done I will upload the designs for both views and we could use this to decide what we think is best for the users needs.

jh-RLI:
Yea, thats what I meant there is a lot to visualize :D 
also don't forget that this PR is more about the [schema page](https://openenergy-platform.org/dataedit/schemas) (a schema is a Topic or Data Category, the therm schema should soon be renamed). They also are the collection of either all tables in case of schema "model_draft" or a subset from all tables like in the case of all other schemas (like climate, ...). So we might want to visualize the connection between the schema and the tables within this schema. **But more important on this page is to visually separate the "model_draft" schema from the other schemas.**

the PR about the table [list page](https://openenergy-platform.org/dataedit/view/model_draft) is this one:
https://github.com/OpenEnergyPlatform/oeplatform/pull/942

PR #950, opened 2022-04-20
fix & simplify filter in schema and tables

### LINKED ISSUE #936: Filter functionality under schema shows wrong results
When filtering under https://openenergy-platform.org/dataedit, clicking a light blue tag (scenario name it seems) and filtering shows 196 tables in scenario and 776 in model draft (in case there are some in model draft). This also holds when clicking two tags. 
Two examples pasted in below. These numbers seem the total number of tables in the corresponding schemas. 
This seems a bug and I think at one point in the past this was solved already as I recall obtaining correct results (but before reorganisation of tags, and I cannnot locate the closed issue). 

related to #712  

grafik

wingechr:
* simplified filter for tables & schemas
* filter shows correct number of tables in schema view
* fixes #936

### DIFF
--- a/dataedit/templates/dataedit/dataedit_schemalist.html
+++ b/dataedit/templates/dataedit/dataedit_schemalist.html
@@ -16,7 +16,7 @@
         </tr>
     </thead>
     <tbody>
-        {% for schema, description, tables, tag_ids in schemas %}
+        {% for schema, description, tables in schemas %}
             <tr onclick="window.location='/dataedit/view/{{ schema }}{% if request.GET %}?{{ request.GET.urlencode }}{% endif %}'">
                 <td>{{schema}}</td>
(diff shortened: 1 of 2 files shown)


jh-RLI (commented):
The changes work!. I just have 2 thoughts about it:
- I think it would be useful to show the currently selected tags extra (in addition to the checkmark that is shown in the "big" tag list), e.g. here:
image

- Would it be good if the user always gets all tables displayed that contain the selected tag(s). So no "and" but an "or" filter for tags. The tags search could then be expliziet again via the combination with the text search. (Just an idea, in general it makes sense that the user gets the result that is expected e.g. filter tag a, b and only get tables that contain a combination of tag a and b).

Translated with www.DeepL.com/Translator (free version)

Issue #901, opened 2022-02-24
Tutorials: improvement ideas for markdown editor capabilities


han-f:
While working on WP 3 an adding tutorials, I ran across some limitations from which I think solving those can help make (text) tutorials look better, I'll list them here for consideration. I also learned there are different "flavours" of markdown that have different functionalities. Which "flavour" do we currently use? 

- Possibility to scale images (maybe just me, but didn't figure it out and since raw html not allowed, didn't try this)
- Links that open in new tabs (helps showing something elsewhere while still being able to read the tutorial in parallel)
- raw html for folding table of contents (see #825  )
- text anchors (see: #778)
- blockquotes (see: #777)

cc @Anne-Oeko

han-f:
Get in touch with @han-f

Issue #825, opened 2021-12-07
allow raw html in markdown editor for tutorials


han-f:
We discussed at the SIROP project meeting Dec 2/3 2021, that it would be helpful to allow for raw html in the markdown editor. this would allow better structuring, specifically of the [FAQ tutorial](https://openenergy-platform.org/tutorials/17/) that may be overwhelming if we add further questions and answers. 

This is related to #778 which deals with improving markdown capabilities, but which also indicates , that raw html is not  allowed for security. I took from the project meeting that the advantages of raw html may outweigh the seemingly small security issues?

cc @Anne-Oeko

han-f:
As we are progressing with WP3 it would be great to hear about how this is coming along?

han-f:
Get in touch with @han-f

Issue #778, opened 2021-08-05
Sections in tutorials can not be linked to


wingechr:
## problem

* it would be really helpful for the tutorials to be able to link to anchors in the text
* but, related to #657:  we cannot/will not allow raw html in markdown for security reasons

## proposed solution

after the markdown -> html parser has finished, add code that creates automatic textanchors for all headers so that tehy can be linked to

@han-f, @Anne-Oeko

johannwagner:
I actually fixed this in some of the tutorial rework branches. There is a permalink link next to each heading, that can be linked.

wingechr:
That's great, I did not realize that, my bad.

wingechr:
Sorry, i closed this too early.

The permalinks only work for the tutorials coming from github jupyter notebooks, not from the markdown tutorials.
Also, maybe fix #777 in the process

han-f:
As we are progressing with WP3 it would be great to hear about how this is coming along?

han-f:
Get in touch with @han-f

Issue #777, opened 2021-08-04
Tutorial editor: Formatting a blockquote doesn't render


han-f:
I am working on a FAQ Tutorial (https://openenergy-platform.org/tutorials/17) and want to format the questions in blockquote as shown here: https://www.markdownguide.org/basic-syntax/#blockquotes-1

Neither manually using the ">" nor the corresponding symbol (") from the editor menu does yields the text to be formatted as a blockquote. 

Can this be fixed?

han-f:
Get in touch with @han-f

PR #951, opened 2022-04-21
Fix/markdown html 901

wingechr:
#901, 

* image scaling can be done by using raw html + bootstrap classes or css (#901)
* links that open in new tabs (#901)
* raw html (#825)
* text anchors (using raw html OR auto generated ids for headers) (#778)
* blockquotes (#777)


jh-RLI (commented):
Great changes! Just a note: It's not self-explanatory how to link to an anchor with the auto-generated ID, e.g. "First Title" becomes "first-title" and you need to know how to create the url to that anchor, e.g. `[a](https://openenergy-platform.org/tutorials/xy#first-title)` .
Since the user is in edit mode, they can't just copy the url and then paste the #anchor at the end, they even have to delete the /edit/. Also, now a new tab is always opened to navigate to the anchor. This might irritate the user as well. Nevertheless, the functionality is much better than before.

PR #953, opened 2022-04-25
fix notebook tutorials (+formatting)

wingechr:


jh-RLI (commented):
Maybe we should change the print statements to logging? But it does what it is supposed to do anyway. The only thing missing to make it work is to remove line 61 `template_file="openenergyplatform",`

PR #978, opened 2022-05-25
Improve dataedit filter sidebar (improve tag search and tooltips)

### LINKED ISSUE #965: Improve usability of tags when using the data filters
- Show the currently selected Tags (regarding data filters):
I think it would be useful to show the currently selected tags in addition to the checkmark that is shown in the "big" tag list:
image

- We should consider changing the presentation of tags. 
  - @Ludee suggested introducing categories which include a subset of tags. This would be useful, but there are still too many tags IMO. 
  - There are several technical/functional solution we could take advantage from. For Example, we could introduce a dropdown for all tags or tag category that provides a search bar at the top. This would also fix the first point, as the presentation of tags changes.
  For example, there are these jQuery plugins for bootstrap 4: [bootstrap-select](https://developer.snapappointments.com/bootstrap-select/examples/#live-search_1) , [select2](https://select2.org/getting-started/basic-usage)

jh-RLI:
- general restructure of the filter sidebar.
- add the jquery bootstrap plugin bootstrap-select and implement select module with life search

closes #965

### DIFF
--- a/dataedit/templates/dataedit/filter.html
+++ b/dataedit/templates/dataedit/filter.html
@@ -54,6 +54,7 @@
         <li class="list-group-item">You can create <a href="/dataedit/tags">new Tags</a>.</li>
     </ul>
 </div>
+<!-- Modal content used in "tooltip" next to Filter section -->
 <div class="modal fade" id="exampleModalCenter" tabindex="-1" role="dialog" aria-labelledby="exampleModalCenterTitle"
     aria-hidden="true">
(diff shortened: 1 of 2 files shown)


jh-RLI:
This is the current result, background color is bad, but I wanted to test it out - it seems like it is not worth it. We could also include categories quite easy @Ludee @wingechr :
tag_search

wingechr (approved):

Issue #943, opened 2022-04-05
Improve usability upload Wizzard


chrwm:
The upload Wizzard on the OEP is a great feature to upload single/few tables. However, having used it recently, I want to propose some improvement suggestions regarding the workflow. 

1. Introduce option to upload one vs many tables (from csv) (after each table and finished metadata -> create table)
2.  Switch the workflow order and make "Upload csv" the first step
4. Then transfer the table headers from the csv-file automatically to the next step "Create Table" to avoid manual copy&paste of table headers
5. Set the checkbox "nullable"  by default to `True`
6. ? Open question: Is it possible to auto. set the datatype for each column based on the CSV value?

Additional: Function to save the current progress when creating one or multible tables locally. (Without creating the table on Database)

pajot:
This is an excellent idea. Having to define a table first without any helper tools is a show-stopper for most users.

What do we have to do to make this happen?

PR #988, opened 2022-06-02
Rework upload wizard #943

steull:
As mentioned in #943, when creating metadata by using the wizard, the 'upload csv' section should be before 'create table'.


jh-RLI:
i forgot to mention that it is better to create a draft pull request. I think you can change this afterwards. ATM this is a "normal" pull request but as this work will not be included for some time we should start off with a draft PR :)

jh-RLI:
Will be done later on

Issue #997, opened 2022-06-13
Broken link in `About OEP` page


Ludee:
The link on the first image is not closed.

PR #998, opened 2022-06-13
Close hyperlink on image #997

Ludee:


jh-RLI (approved):
Good catch. We should Include a html Linter so bugs like this will be detected automtically.

PR #1004, opened 2022-06-16
Add tag color indicator to tag filter dropdown

jh-RLI:
Add a fontawesome circle icon to each select option and style the icon using inline css and the tag color.


wingechr:
works here, i like it. 
@jh-RLI I added 2 small changes:

* add a "Deselect All" button (actually, a "Select All" is also added, but I hide it via css)
* if multiple tags are selected: only show tag count, otherwise it gets crowded rather quickly or tags cannot be shown, especially on a smaller laptop screen)

grafik


What do you think?

jh-RLI:
I like it. Maybe we should make the deselect all button more prominent? Not sure what an user thinks about this :) To me this looks great as it is.

wingechr:
yes, i agree, there should be an area displaying the currently selected tags
also, once you click apply, the page is reloaded and the selection vanishes, that should be fixed too

wingechr:
@jh-RLI

* I added code to keep the tag filter populated on page reload
* I also REVERTED back to showing all selected tags (instead of count) because
  * i made the div wider (same width as text box)
  * the tooltip automatically shows all selected values
  * if we going to move the filter from the sidebar to the main section later, we will have probably enough space

grafik

wingechr (approved):

PR #1007, opened 2022-06-27
Update filter description

jh-RLI:
The description of filters (text and tags) for data was revised some time ago, but the changes were lost in the process. This PR introduces an update of the text.


wingechr (approved):
looks good here

PR #1022, opened 2022-07-30
Fasttrack minor changes

wingechr:
as discussed: some small fixes on the fasttrack branch


jh-RLI (approved):
Great changes! I missed the 'static/'. Did I understand you correctly that you also solved the factsheet editing problem? At least I was not able to reproduce the error that made editing a very bad experience, since the user had to clear the cache and reload the page for each change.

wingechr:
not sure what problem you refer to, but maybe. i just fixed the order of some of the javascript code, because it requires jquery to be loaded first.

PR #1023, opened 2022-08-09
added tutoral category overview that is sticky

### LINKED ISSUE #826: Restructuring tutorial page to hold capacity building packages separately
We discussed at the SIROP project meeting Dec 2/3 2021, that it may be a good idea to integrate capacity building packages using the (improved) markdown capabilities and add them to the [tutorial section](https://openenergy-platform.org/tutorials/). As they provide more context than tutorials they (and the FAQ) should stand out from the tutorials. I attach a first scribble on how this could look like. Please note that this is only a scribble for brainstorming further and also the texts are only first ideas to provide a bit context.

sirop-capacityandtutorials

cc @Anne-Oeko

### LINKED ISSUE #1010: Add "capacity building module" to drop-down menu in tutorials
We want to add this specific category to the drop-down menu in the tutorial editor, so all capacity building modules can be displayed on a separate section on the tutorial page (#826). 

cc @jh-RLI

wingechr:
first simple implementation of new tutorial category `overview` that will be sticky on top.
we still have to determine later how permissions will be restricted later.

closes #1010
closes #826


jh-RLI:
Do you want me to review this already?

wingechr:
sure, it's a very minimalistic implementation, but enough so that it works. let's discuss in today's dev meeting

jh-RLI (commented):
Works for me, the overview tuorials show up below the overview heading. 
I think sticky is a bit misleading (at least for me) I understand it as an element that is also visible when you scroll down. It sticks to the top of the screen for example. Or have I misunderstood what you mean here.

jh-RLI:
@wingechr what do you thin about this 
image

PR #1030, opened 2022-08-12
static files instead of external resources and remove deprecated file 

wingechr:


jh-RLI (commented):
@wingechr I get this when viewing the dataedit/schema and table pages. 
image

If i then open the tag select dropdown there are no tags visible for a few seconds and it throws another error. Everything works fine after that first hickup. 

I think this eats a lot of spagetti :) Great work.

wingechr:
the source mapping is only for development and not needed,so i removed it from the js file

jh-RLI (approved):
Now everything works as expected.

jh-RLI:
I removed a deprecated file and just added it to the fasttrack branch. The file contained deprecated settings and is not related to the static files.

PR #1032, opened 2022-08-12
when creating table, auto create and check existence of numerical id …

### LINKED ISSUE #739: I get error "column id does not exist" on insert only
This seems to be intended behaviour, but it is not consistent.

* If tables need an id column, the create table request should already fail if the column does not exist
* Furthermore, the insert fails only via the API request. It is still possible to do it via the advanced api insert call.
* This is also not well documented

We need to decide if this should be the intended behaviour or not

* if not: remove the restriction
* if yes: make it consistent for all API calls

I can write the patches, but we first need to decide what the intended behaviour is supposed to be

wingechr:
closes #739

### DIFF
--- a/api/actions.py
+++ b/api/actions.py
@@ -749,12 +749,12 @@ def assert_valid_table_name(table):
                        "and start with a letter.")
 
 
-def table_create(schema, table, columns, constraints_definitions, cursor, table_metadata=None):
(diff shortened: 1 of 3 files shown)


wingechr:
also raises an API error when user provides duplicate column names

jh-RLI (approved):
Very nice! I think this will add a lot of value. I will also adopt your testing workflow.

Issue #658, opened 2020-10-29
Missing integration of review process in platform


christian-rli:
The review process is too complicated, especially if to attract volunteers. Significant parts of it can be automated. All of that should be integrated into the oep. If it's easy and maybe even fun to review, it'll make life easier for reviewers and maybe even attract volunteers.

jh-RLI:
This new open peer review process is actively implemented in the platform.
![image](https://user-images.githubusercontent.com/38939526/236575877-98c5d401-0062-4eab-85ba-bffaa10ca331.png)

jh-RLI:
The first version is well on its way. The review process is already implemented in the platform and is being actively further developed. See https://github.com/orgs/OpenEnergyPlatform/projects/15/views/1?layout=board

PR #1033, opened 2022-08-15
Draft to develop Open-Peer-Review-Process on OEP

steull:
Development of #658 and #1149


jh-RLI:
The CI fails for python 3.8 as the type hints have not been added by default back then - this changed in python 3.9:
https://peps.python.org/pep-0585/#implementation

We would need a workaround like importing `from __future__ import annotations` in omi - or we should drop tests for python 3.8 and make 3.9 the new minimum version - we should discuss in an upcoming oep-dev meeting.

henhuy:
Here my suggestions for further code improvements (surely to be discussed)
In frontend:
- do not hardcode categories in template(see backend recommndation for metadata)
- JS functions are now declared globally (this was my fault - saw that this is handled differently in other views) - should be moved into config function - to do so, all `on_click` in template must be refactored into JS click events in config function!
- improve submitting of feld review (ok should call submit directly, suggestion and reject should "open" dialog for field suggestin and/or comment)

In backend:
- Do not put metadata as is into context, but preprocess it to deliver dict with categories and fields (much simpler to handle in template!)
- store reviewer in model field (not in review JSON and not for each field, but only once)
- store timestamp of review once in model field
- do not store version in review JSON (can be calculated from model)

Hope this helps!

jh-RLI:
@christian-rli I genrated (`python manage.py makemigrations`)and added the missing migration. Now the CI Tests are running again. Please pull my changes :)

PR #1034, opened 2022-08-15
Feature/command check dead links

wingechr:


wingechr:
example call:

```
python manage.py check_links --no-external http://localhost:8000
```

PR #1039, opened 2022-08-19
tag filter: update on change, remove apply button

### LINKED ISSUE #832: Tag Filter should return instant results (asynchronous requests/response)
Users expect that tables will directly be filted upon a klick on the tags in the side. The Apply Filter button is overlooked. Make and implement a concept for fixing this and while at it, maybe also create a cleaner display of the tags on the side (as called for in #857 ).

wingechr:
closes #832

### DIFF
--- a/dataedit/templates/dataedit/filter.html
+++ b/dataedit/templates/dataedit/filter.html
@@ -25,17 +25,17 @@
         <i class="fas fa-info-circle" style="color: #3B77A0;"></i>
     </button>
 </div>
-<form action="{{ request.path }}" method="get">
+<form id="form-filter" action="{{ request.path }}" method="get">
     <h5>Text</h5>
     <div class="searchbox_container small">
-        <input type="text" name="query" placeholder="Search" value="{% firstof query '' %}" class="form-control" />
+        <input id="search-filter" type="text" name="query" placeholder="Search" value="{% firstof query '' %}" class="form-control" />
     </div>
     {% load dataedit.taghandler %}
     {% get_tags None None None as all_tags %}
     <h5>Tags</h5>    
     <div id="tags">
         <div style="display: flex; flex-flow: wrap; margin-bottom: 1em;">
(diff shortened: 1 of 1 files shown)


jh-RLI (commented):
On my side the filter is applied on the first click on a tag, is this the intended behavior? I think if a user wants to filter by multiple tags, they will be confused with the behavior of the tag filter dropdown because it always closes on "select tag" because the page reloads. This makes the experience a little crazy, what do you think? 

Maybe we should avoid reloading the page altogether and update the table list using ajax call?

wingechr:
i discussed the user experience with han-f in #832. the problem with ajax is that we then cannot have the url reflect the current selection, which (and i agree) is a very useful feature for users. i see multiple options:

A) use the current crazy implementation (con: multiple reloads, pro: with each selection, user sees changes in selected tables)

B) switch to ajax (pro: immediate feedback. con: url not up to date)

C) only reload when user closes the selection box (without ajax)

D) hybrid of B and C: reload page when user is done (i.e. closes the selection box), but update with ajax during select/deselect)



i guess D is the best but also the most complicated

jh-RLI:
I think if we do not implement the complicated solution D I find the previous solution (open dropdown, select tags, click apply) better, because it is understandable for the user and already implemented. This would also be a solution from my point of view. Do not forget that the issue still assumes the old tags, at that time there was not yet the dropdown.

Also I see two more options here:

E) use Ajax and provide a shareable URL to share the result view. The URL could be constructed on the server based one the filter request and then made available via some field or button.

F) I should test this first, but it might be possible to update the URL after the successful Ajax call:
- https://stackoverflow.com/questions/824349/how-do-i-modify-the-url-without-reloading-the-page

wingechr:
nice i did not know about hostory.pushState - i will try to implement F

wingechr:
or maybe not, there is a lot of things that would need to change. let's discuss tomorrow

PR #1040, opened 2022-08-23
Feature/migrate to bs5 1027

### LINKED ISSUE #1027: Update to bootstrap 5
It would be nice to update from bootstrap 4 to 5. 

We would have to upgrade the PyPi package `django-bootstrap4` to https://pypi.org/project/django-bootstrap5/
and also review all bootstrap related code and files: https://github.com/OpenEnergyPlatform/oeplatform/search?q=bootstrap

bmlancien:
Still in progress...

I installed the new django-bootstrap version [this way](https://django-bootstrap5.readthedocs.io/en/latest/installation.html), but then later I got the error mentioned [here](https://stackoverflow.com/questions/72276616/django-modulenotfounderror-no-module-named-bootstrap5), so I tried it this way and it worked fine.

For some reason, I wasn't able to use the minified versions of the Bootstrap JS, so I included the [CDN](https://getbootstrap.com/docs/5.2/getting-started/download/) too as links to make it work properly. But I guess you'd prefer to use self hosted files than relying on CDN links? Maybe the paths are wrong.

Also, there is no jQuery in Bootstrap 5 anymore but there is still some jQuery in the OEP code. We need to check this too.

I'm still going through the _breaking_ tags from https://getbootstrap.com/docs/5.0/migration/ but most of them are done.

Some styles on the platform are now different, e.g. the blue color. I also need to check/change that.

close #1027


wingechr:
hi @bmlancien. 

* i tried to fix the bootstrap dependencies.
* i used the theming to create `base/static/css/bootstrap.min.css` like this
  - navigate into the `theming` directory
  - locally clone bootstrap: `git clone --branch v5.2.0 https://github.com/twbs/bootstrap.git`
  - build css: `sass --no-source-map --style=compressed oepstrap.scss ../base/static/css/bootstrap.min.css`

* some pages look a little crappy now, for instances the login page (`/user/login`):

grafik

wingechr:
also please note: the first commit 5380063 is a whole bunch of changes from merging dev into the branch to include all the reformatting of all of the code

bmlancien:
Thanks @wingechr for your help! I'm gonna check if there are any layout bugs. There must be a few ones apart from the login page you just mentioned. Then I will push the changes.

bmlancien:
I had a look at the website and fixed a few things about the layout that were broken. There are still a few very minors things that I see that are different now (like another blue color for the login button or a few pixels more on the footer height) but nothing very important (also because the website design should be updated soon). But could someone have another look to see if they discover layout issues?

I also have a doubt regarding the forms. BS5 now includes `.form-label`as a CSS class for form labels. I added it to the HTML templates and I also tried to do it to the JS files that create/inject HTML forms, e.g. in `dataedit/static/dataedit/metadata.js`. I'm just not sure if I maybe have missed other label mentions that are not so obvious to find inside the JS files ?

Something I also changed was the use of `@use` and `@forward` instead of `@import` for SCSS files because as mentioned [here](https://stackoverflow.com/questions/62757419/whats-the-difference-between-import-and-use-scss-rules): 

> @import will be deprecated in favor of @use and @forward, and support [will be dropped] by October 2022 at the latest.

That means I had to change a bit the structure of the SCSS files. It's actually a bit problematic with BS because they don't offer a ready-made solution to import their variables with `@use`. So the best solution I had found working on another project was to add `theming/scss/base/_index.scss` and from there "forward" the intact BS variables as well as the BS variables updated by us so we can have access to them inside the other SCSS files such as `theming/scss/components/_tables.scss`. This way it is possible to list the SCSS files inside `theming/oepstrap.scss` with `@use`(there will be more files in the future).

jh-RLI:
I will also have a look at the updated layout.

jh-RLI:
I think this is great. Most of the layout seems correct to me. 

You are right about the form, but in gerneral they work. When looking at the create/add tutorial page all dropdowns look like this the other input fields look good.
image

jh-RLI:
We should be testing this on TOEP soon since it is public to other testers and includes the ontology app, but I don't think the change will affect it.

jh-RLI (approved):
I have tested the user interface and it looks good for the most part. If it's easy to change the buggy forms, please add this to this PR. Otherwise we can merge it and fix the rest later?

jh-RLI:
I tested it again and the broken form was just some bug on my local setup. So it seems like everything is working.

Issue #735, opened 2021-02-26
Improving the implementation of OEM with regard to data models with multiple tables / improve datapackage support


jh-RLI:
In the course of the development of the oedatamodel and also e.g. in the current review of the oemof_b3 dataset it becomes apparent that there is no good solution on how to append the same metadata to several related tables (as far as i know). This results, among other things, in the problem that, for example, C(R)UD methods are difficult to use from the user's point of view. In this case, each table must be handled individually in order to distribute a change. 
To illustrate: A user wants to upload a dataset that is spread over several tables (e.g. 15). After uploading, metadata is appended to the tables. Now a new source is to be documented in the metadata. The user knows the OEP and uses the web interface to edit the metadata. Unfortunately he has to update the metadata for each of the tables individually :( 

From a developer's point of view, this may not be a problem, as it is always easy to replace the metadata "quick and dirty" on all tables concerned. However, my main concern here is usability.

I have created an overview image for the use case (a possible solution IMO). I think more work will be needed to implement and display data models that include multiple tables on the OEP website. 

![CRUD_md_on_multiple_tables_same_dataset](https://user-images.githubusercontent.com/38939526/109312277-24c78a00-7847-11eb-9884-471f70e7beca.png)

Ludee:
This means metadata will not be stored as a comment anymore but will be moved to a central "metadata table". This will be a major rebuilt of the metadata structure and will cause many changes in different modules!?

jh-RLI:
This could be the case. But it is also possible to keep the metadata as comment on table and update it when changes occur. The table with the metadata serves here only as a central resource in which changes are detected and then distributed (as comment on table).

jh-RLI:
#1059 a new method for saving and reading oemeta data to and from the database is introduced. This solves part of this issue. The next step is the implementation of a data collection store consisting of several tables (which are technically not connected, but belong to the same area/project). In addition, tables created as relational data models should also be visualized on the OEP. If we want to follow the idea presented here, we would also have to implement some functions to make data collections smoothly compatible with frictionless data packages. It is therefore necessary to generate a datapackage.json file from the metadata, which describes all resources of a datapackage.

PR #1059, opened 2022-09-08
improve table.oemetadata loading time #735

jh-RLI:
- [x] Store oemetadata in dataedit.models.tables oemetadata (JSONB) field (update api post and get still reads from SQL comment on table)
- [x] Add functionality to get correct oemetadata key order as JSONB does not preserver order.
- [x] Adapt metadata and MetaWidget functionality

Addtional:
- [x] fixes popover in bootstrap5
- [x] Add Tests


jh-RLI:
@wingechr Please review this as soon as i mark as normal Pull Request (not draft). 

At the moment, I would be interested in your opinion to extend [this table](https://github.com/OpenEnergyPlatform/oeplatform/blob/c4f61fbcf29016c80644081a00d30223759520e5/dataedit/models.py#L40-L56) and to store the metadata string in a JSON field there instead of creating a new table for oemetadata. My question is, do you think we will keep this table after we resolved the schemas.

Regarding the schema solution, we could reuse some django tables that store info on all existing tables and the schema they are stored in. The [schema table](https://github.com/OpenEnergyPlatform/oeplatform/blob/c4f61fbcf29016c80644081a00d30223759520e5/dataedit/models.py#L35-L37) (rename to topics) in combination with the [tables ](https://github.com/OpenEnergyPlatform/oeplatform/blob/c4f61fbcf29016c80644081a00d30223759520e5/dataedit/models.py#L40-L56) table to implement topics (tag tables with corresponding topic). 

I have spoken with ludwig and we are definitely in favor of keeping the topics, since a lot of work went into the elaboration back then.

wingechr:
Yes, definitely use this the Table object. This is the django representation of an oedb table and we will definitely keep it. creating a new model class  is not necesary because it's one-to-one relation.

wingechr:
I'll review it as soon as possible

jh-RLI:
Thanks a lot, I will eventually look again and maybe refactor some code so it is more organized. In gernal it works. I'm not so fan of keeping the SQL comment on table and the DB json field together (as a store for the metadatan) but did it that way now, as it was initially my plan to keep the old functionality.

wingechr:
in addition to some black formatting, i changed (simplified?) the way the nested parsing of the metadata (try 1.5, then 1.4) is done in the `api/actions.py`.

To me, this seems more readable now, but we should look at this together

jh-RLI:
Yes! i think it is more readable, i wasn't able to come up with a better way for the exception block and looked deeper into exceptions - based on what i learned there your solution is simple and great :D .

wingechr (approved):

jh-RLI:
@wingechr what do you think, should we merge this already? I've started revising the migration script (Martin's outdated one) and added it to the new Django migration. I could add this, but this PR is getting kinda big.

jh-RLI:
I think i will add the migration here.

jh-RLI:
I will merge this PR now and add the migration stuff to a new one.

PR #1060, opened 2022-09-12
MetaEdit Standalone readonly #1054

### LINKED ISSUE #1054: MetaEdit Standalone some fields are readonly and other minor bugs
The standalone MetaEdit should allow the user to enter field names and types. They are read-only as they are automatically filled in if a table has already been created. Since no table is created when using the standalone version, the fields must be filled in by the user. 

image

Additionaly the header shows:
image
This should be changed to something more understandable.

steull:
closes #1054


jh-RLI (approved):
It may not be necessary to use the else section, since the default value is false. But it might be better to be explicit.

Issue #1062, opened 2022-09-14
enforce naming conventions for tables and columns


wingechr:
see also https://github.com/OpenEnergyPlatform/oem2orm/issues/16

jh-RLI:
We once started collected naming convetions here: https://github.com/OpenEnergyPlatform/data-preprocessing/wiki#naming-conventions.
Perhaps it would be worth keeping a list of the checks that have been implemented in the OEP API?

wingechr:
both tables and column names now must adhere to the regexp '^[a-z][a-z0-9_]{0,MAX_LEN-1}$', MAX_LEN currently being 50.

The length limit is a little shorter than the postgres restrictions (63 for tables, 59 for columns), because of the versioned tables that add suffices (like `_edit`) to the table names

chrwm:
> both tables and column names now must adhere to the regexp ^[a-z][a-z0-9_]{0,MAX_IDENTIFIER_LENGTH-1}$, MAX_IDENTIFIER_LENGTH currently being 50.

@wingechr is there a reason to put a limit of 50 chars on the `resources-name`? I already ran into problems regarding this and had to rename tables. Thus, I'd vote to increase the the limit or to omit it.

jh-RLI:
This is a technical limitation due to the Postgresql database - if there is no better solution than [this] (https://stackoverflow.com/questions/8213127/maximum-characters-in-labels-table-names-columns-etc), unfortunately we cannot change this.

chrwm:
If there is no better solution, I'd suggest we add a section on the OEP nearby the upload Wizzard that highlights all the postgresql naming conventions. 
Knowing them in advance and taking them into account is more resource efficient than correcting them later.

jh-RLI:
Then let's make a list in this or a new issue and collect all things there. The list should represent the conventions in a readable form. Then we can add them to the mockup Bryan created.

chrwm:
Oki I'll start here, as the issue title seems appropriate

chrwm:
Do's and Don't apply for 
1. tables names
2. table headers

#### Do's
* use ASCII characters only
* use lower case only 
* use underscores

#### Don'ts

* table name must not exceed maximal character limit = 50
* no points
* no commas
* no spaces
* no special characters
* avoid dates
* if dates are used, then without `-`! E.g.: 2022-10-21 will through an error. Use underscore instead!
* don't start parameter name with a number

wingechr:
[...]

PR #1065, opened 2022-09-18
enforce column name format as well as table names

wingechr:


wingechr:
closes #1062

jh-RLI (approved):
Works for me! Because you provided tests, I just tested the table creation via website. Also added info about the length restriction of 50 characters to the table create tooltips.

PR #1111, opened 2022-11-16
PK constraint is skipped if PK is set in column

### LINKED ISSUE #1110: Multiple definitions of primary key error when defining PK on table column
See https://github.com/OpenEnergyPlatform/oedialect/issues/43 for reference

I will add a small PR to address this.

henhuy:
Fixes #1110

This PR checks whether PK constraint is already present as PK in column.
If PK (pointing to same column) is already present, no error is raised and PK constraint is not added to constraints.

Could you review @wingechr ?
Attention: Not tested, as I have no OEPlatform setup!


jh-RLI (approved):
Works on my local setup.

PR #1113, opened 2022-11-16
Update tutorials to bootstrap 5

MGlauer:
Bootstrap 5 uses a different property for data references


jh-RLI:
We removed the tutroials app ... and migrated to mkdocs.

PR #1114, opened 2022-11-16
Fix tutorial accordions

### LINKED ISSUE #1112: Tutorials (CBM) are no longer extensible (UI).
This is probably due to the new Bootstrap 5 HTML tags.
https://openenergy-platform.org/tutorials/#tut-training-start

image

bmlancien:
@jh-RLI I fixed the broken accordions in the tutorial sections 02 and 03. I actually had to change a bit the HTML structure, so it was not just an easy fix with a missing class. Therefore it would be good if someone could check on that to see if everything is still there. I saved the older tutorial versions because I don't know where/if they are saved somewhere?

Regarding the arrow pointing down on section 01, 02 and 03, actually they were not built as accordions but as buttons to go back to the top of the page. I found it quite misleading, I also thought it was supposed to be accordions. So I just removed them because I don't think we need them. If we really want that go-back-to-top button, we should think of another solution.

close #1112

### DIFF
--- a/tutorials/templates/list.html
+++ b/tutorials/templates/list.html
@@ -27,14 +27,14 @@
 
 <div id="tutorials-overview">
 {% for tutorial in tutorials_overview %}
-  <h3 class="my-2"><a href="#tut-training-start" title="back to top"><i class="mr-2 fa fa-caret-up"></i></a>{{ tutorial.title }}</h3>
+  <h3 class="mt-5 mb-3">{{ tutorial.title }}</h3>
   <div class="tutorial-container">
     {{ tutorial.html | safe }}
   </div>
 {% endfor %}
 </div><!-- end of overview tutorials -->
 
-<h2 class="my-5">Tutorials by category</h2>
+<h2 class="mt-5 mb-3">Tutorials by category</h2>
 
 <table class="table table-bordered">
     <thead class="table-dark">


jh-RLI:
I just saw that there is another fix for the tutorials in PR #1113 but for the backend - maybe that's what was missing to just change the class - we can check tomorrow if you want. Good thing you saved the tutorials, they are saved but when you change them the old ones are overwritten. Maybe the new changes are even better than what was there before ?

bmlancien:
@jh-RLI Sure we can check this. I took the occasion to make a few styling changes, but we can use the old ones if preferred.

jh-RLI:
Yes, I think it looks great, but those who worked on the old tutorials created some reusable HTML code blocks for non technical tutorial creators - so I'm afraid I can't decide on my own if we want to use your updates or if that will mess up the workflow they designed :) 

But the good thing is that we have your updates now and the old ones too. Then we can present that soon  :)

han-f:
We would indeed require the reusable HTML to still work, as we do have partners who want to start building capacity building modules using the template: https://openenergy-platform.org/tutorials/25/ 

I am ok with another version of a back-to-top button, but would prefer the accordions to be coloured as the colouring had some meaning and helped distinguish visually across the page. 

cc @wingechr

han-f:
fixed the colouring of the accordions.

jh-RLI:
I think if we use the updated style, we could also update the template for the capacity building modules. Then it would be fine. We should harmonize this soonish. 
The style that Bryans is introducing is IMO a bit more modern. Since you already added the colors, I would vote to update the template according to his design. 
There is one more modification ](https://github.com/OpenEnergyPlatform/oeplatform/pull/1113) that is relevant to the backend, I will merge this and release it as a patch in the next two weeks.  This would also require a very small update to the template again. 
Do you agree @han-f

PR #1128, opened 2022-12-07
Fixes conda env and instructions

### LINKED ISSUE #1127: Fix conda deps and instructions in README
`environment.yml` does not include `conda-forge` channel, pip as dependency and shapely.
Also, conda install instructions does not use `environment.yml`
Also, old python version is set (uses `python=3.6`).

henhuy:
Closes #1127


jh-RLI (approved):
Thanks for the update. This part has not been touched for a few years.

We still have some inconsistencies regarding the Python version we are using:
- The production server is running Python 3.6 
- the ci and the readme statements are on newer versions (Py 3.8 and 3.9)

jh-RLI:
@henhuy sorry for making the ci fail here :) i wanted to test if it fails on every branch .... seems like something changed in the background that breaks our current testing workflow using github actions.

Issue #1141, opened 2023-01-06
OEP Page Content - Factsheets Overview needs update


christian-rli:
As suggested in #1126 this is a single issue for updating the content of the Factsheets Overview Page. The current content reads as follows:

Factsheets Overview

The Open Energy Platform hosts a series of factsheets – currently we offer [Model Factsheets](https://openenergy-platform.org/factsheets/models/), [Framework Factsheets](https://openenergy-platform.org/factsheets/frameworks/) and [Scenario Factsheets](https://openenergy-platform.org/factsheets/scenarios/) that help to learn more about the context of the data you can find in the database. The Factsheets are a standardized collection and presentation of information about modelling software used in climate and energy system modelling. The use of interactive fields and pre-defined responses is designed to make it easy to insert data and filter for existing entries. For example, you may want to know more about the background of some of the data you found than the metadata attached directly to it can provide. Then factsheets are a helpful place to visit.

Specifically, our current Scenario Factsheets provide some dedicated information about specific scenarios for which projection data is available on the OEP. Please note, however, that their format was developed some time ago and they will undergo substantial updates. In the future, these scenario factsheets will also be succeeded by [Study Factsheets (alpha)](https://openenergy-platform.org/factsheets/rdf/study/) that will hold more functionality. These improved factsheets will be based on RDF and the [Open Energy Knowledge Graph (OEKG)](https://github.com/OpenEnergyPlatform/oekg) which is currently under heavy development. Please note that this is just a preview and not a fully functional feature yet.

These study factsheets will inform about scenario contexts and enable you to explore scenario bundles. Scenario bundles tie together various information about a scenario available on the OEP. This includes information on the scenario projection data available, the frameworks and models applied, and the study context. The study factsheets will make these relations visible and usable. And not only that: the study factsheets will be based on the concepts available in the Open Energy Ontology (OEO).

Anyone is invited to contribute to development as this is a community-driven process. Most of the development takes place on GitHub. We have some [helpful guidelines](https://github.com/OpenEnergyPlatform/oeplatform/blob/develop/CONTRIBUTING.md) on the workflow providing detailed information. [Welcome! We invite you to participate](https://github.com/OpenEnergyPlatform/ontology/wiki/Welcome!-How-to-participate).

![Screenshot_Factsheets](https://user-images.githubusercontent.com/42000429/211027233-a0759b90-842f-4584-a8aa-ef954ed6ee9a.png)

jh-RLI:
[...]

PR #1144, opened 2023-01-06
Update the Factsheets Overview Page

christian-rli:
Update the Factsheets Overview Page #1141


jh-RLI:
@christian-rli 
I noticed that some links in the factsheets overview page are hardcoded. Links that link external ressources like `github.com` are okay but links to `openenergy-platform.org` should not be hardcoded. It is better to use the template tag provided by djnago and reference the linked page by an identifier like the name (that is or must be specifyed in the urls.py). See the [documentation](https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#url).

jh-RLI (approved):
I agree, this reads better.

PR #1158, opened 2023-02-01
management command and test to clean sandbox

wingechr:
As discussed:

We need a management command to empty the sandbox schema.
Please review with care,as this deletes data.

### DIFF
--- a/api/tests/test_clear_sandbox.py
+++ b/api/tests/test_clear_sandbox.py
@@ -1,0 +1,34 @@
+from dataedit.management.commands.clear_sandbox import (
+    SANDBOX_SCHEMA,
+    clear_sandbox,
+    get_sandbox_table_names_oedb,
+    get_sandbox_tables_django,
+)
+
+from . import APITestCase
+
+
+class TestCommandClearSandbox(APITestCase):
+    test_schema = SANDBOX_SCHEMA
+
+    def test_clear_sandbox(self):
+
+        # create a test table with data
+        # so that sandbox is not empty
+        self.create_table(
+            table="test_sandbox_table",
+            schema=SANDBOX_SCHEMA,
+            structure={"columns": [{"name": "id", "data_type": "bigint"}]},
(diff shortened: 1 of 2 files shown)


jh-RLI:
@wingechr  I will finish my tests tomorrow, I had some other issues with my dev environment today. So far I created my test table and added some data to it. Just to be sure: is it a good test case to create the same table in the sandbox and in all the other schemas. Then run your command and test if there are any discrepancies in the other schemas... would that be sufficient?

wingechr:
as a local (one time) test, you can create the same tables in sandbox and another schema.
but as a unit test case in the code, the table names should not be the same (because this test will not work in the future).

jh-RLI:
I tested it manully and run the test sucessfully. Never ran into an error. My database also looks fine. But i did not test any special tables or relations. 
Im not expectinga fatal data lose even if there is a cascade invoved as the statement focuses on the sandbox schema. Think we should just give it a try :)

jh-RLI (approved):
see comment above.

jh-RLI:
@wingechr I think @christian-rli could also do the local testing. But in general, I am also confident that this can be merged.

christian-rli:
It took me a while to get a working setup with a functional sandbox schema. I made it now. How do I actually run the command - i.e. which api request? or is it an sqlalchemy thing? django command? I tried to figure out myself, but figured asking would be quicker.

christian-rli:
> It took me a while to get a working setup with a functional sandbox schema, but I made it now. How do I actually run the command, i.e. which api request? or is it an sqlalchemy thing? django command? I tried to figure it out myself, but then figured, asking would be quicker.

I figured it out after all.

christian-rli (approved):
As jh-RLI I ran the command manually, successfully and everything went as expected. Tables in the sandbox were deleted and apparently no other schema was affected. Tables with the same name as tables in the sandbox also survived in other schemas. I give it a go!

PR #1173, opened 2023-03-14
Fix/clear sandbox: also delete tablesin metaschema _sandbox

### LINKED ISSUE #1169: clear sanbox command doesnot remove tables in `_sandbox`

wingechr:
I had assumed they would be removed automatically too, but were not.

closes #1169


jh-RLI (approved):
Works on my side 👍 

I always have somethin :P :
In the change log, we should also document the PR-ID (#PR0000) so that each change is tracable to the implementation.

PR #1179, opened 2023-03-17
Fix/metadata editor

### LINKED ISSUE #1154: oep metadata edit web interface broken
(probably because the json schema that generatesthe form is not up to date?)

grafik

wingechr:
i finally managed to get a migration script (as a management command) for the metadata: `dataedit/management/commands/upgrade_metadata_20230317.py`

* i testet it over ALL existing metadata and i think it works
* the fixed metadata is not only all compliant with the latest omi parser, but also passes the json validation from the schema file.

it's rather long spaghetti code, but there was a lot of fringe cases to consider.
the main reason is that omi still drops a lot of fields it does not recognize, so i had to map them all to the new correct fields to minimize data loss.

you can run `python manage.py upgrade_metadata_20230317` to check if the code works for all tables without changing anything. if there is no exception thrown, you can run `python manage.py upgrade_metadata_20230317 --fix` to actually write the changes into the database.

if you want to just check/test the code on a metadata object, you can use the main function: `fix_metadata(metadata: object, table_name: str) -> object`

closes #1154

### DIFF
--- a/versions/changelogs/current.md
+++ b/versions/changelogs/current.md
@@ -2,4 +2,8 @@
 
 ### Features
 
+- one time migration script to upgrade legacy meta data
+
 ### Bugs
+
+- truncate keywords/tags to 40 characters


wingechr:
another note: there are currently  3 (!) different ways to read metadata

* oemetadata attribute in django table object
* comment attribut in sqlalchemy table object (oedb comment)
* directly reading table comment from oedb

I think i will simplify this and add these commits here later

wingechr:
To help with the review: here is a standalone (without django) version of the code:

[test_upgrade_metadata_20230317.zip](https://github.com/OpenEnergyPlatform/oeplatform/files/11007746/test_upgrade_metadata_20230317.zip)

* you can run it as `python test_upgrade_metadata_20230317.py SCHEMA TABLE`
* it will download the metadata from the oep, apply the fix, and save both for comparison as `SCHEMA.TABLE.meta.orig.json` and `SCHEMA.TABLE.meta.fixed.json`

Here are some changes from the included example (`python test_upgrade_metadata_20230317.py society destatis_zensus_population_per_km2` ):

```diff
"contributors": [
    {
      "comment": "Create table",
      "date": "2017-10-25",
-      "email": "",
-      "name": "Ludee"
+      "title": "Ludee"
    },
```

```diff
"language": [
-    "eng",
-    "ger"
+    "en-US",
+    "de-DE"
]
```

```diff
-  "license": {
-    "copyright": "...",
-    "id": "dl-de/by-2-0",
-    "instruction": "...",
-    "name": "Datenlizenz by-2-0",
-    "url": "www.govdata.de/dl-de/by-2-0",
-    "version": "2.0"

+  "licenses": [
+    {
+      "attribution": "...",
+      "instruction": "...",
+      "name": "Datenlizenz by-2-0",
+      "path": "www.govdata.de/dl-de/by-2-0",
+      "title": "dl-de/by-2-0"

```

```diff
  "resources": [
    {
-      "fields": [
-        {
+      "schema": {
+        "fields": [
+          {


```

jh-RLI:
> 

Yes we can harmoize this once the fix is done :) after that we do not need to keep the metadata as comment on table.

jh-RLI:
I change the metaBuilder: The ID field is now always set to the current URI of the table, otherwise there will be a wrong value after the fix. It's not worth to add another value when the fix is executed .. omi requires any value in the id field, so it's ok as it is.

jh-RLI (approved):
works and tested!

PR #1181, opened 2023-03-22
Feature/extend profile page#1170

jh-RLI:
# Open Tasks
- [x] squash django migrations
- [x] verify  new code will not break anything on the OEP
- [ ] verify that profile image upload is desired as is or we should just enable avatar images. (leagel issues if we store profile images?)

# UI
- we can merge this as it is and implement the ui later on. This is already a lot better then the current profile.


jh-RLI:
@Darynarli I added functionality to get the user tables and I also fixed the bugs. I had to remove the user_id field and the migrations regarding the dataedit models. You should pull my changes and run `python manage.py migrate dataedit 0019` to roll back your database. I also added your changes to requirements.txt, I think it is nice to set the version with ~= ... but how will this be updated, are you using a package to update the dependencies?

Darynarli:
@jh-RLI yes, I use a package to update the dependencies. Must I pull changes on this branch or develop?

jh-RLI:
What is the name of the package - maybe i will use it too. We just need to make sure that we use a consitent way of documenting the packages. 

Yes you have to pull my changes and apply the migration command from my comment.

jh-RLI:
We have to merge this branch soon to use the structre and implement the openpeer related listings of active and open reviews. This will enable collaboration between reviewer and contributor.

jh-RLI:
@wingechr Great :) There is a test for py 3.8 that doesn't seem to be finished.

wingechr:
yea. that's because in develop branch, 3.8 is still in the workflow, but it's not started because it's no longer in this branch. we can ignore that.

But don't merge yet, i still have some issues with the actual code

jh-RLI:
Okay 👍

wingechr:
@Darynarli thanks for your contribution, the user page looks much better already.

* I changed the order of the sub pages so that the settings come first (because I think that's the expectation if a user visits their profile page) - I hope that's ok.

grafik


# Questions / Problems

@Darynarli, @jh-RLI 

There is some issues we need to solve:

* this PR provides a feature to delete the account. We need to decide how this affects the users tables. And we need to create test cases for this to make sure this works as intended (or disable deletion for now?)
* why is there a user name in addition to the name? 
* should we really allow that users can change names? we need test cases for what happens when users change their names to an already existing name (because name is a unique property)
* if users change their email, should this trigger the mail validation again?

Maybe we should discuss these points in a dev meeting?

Cheers

bmlancien:
Hi @wingechr

> I changed the order of the sub pages so that the settings come first (because I think that's the expectation if a user visits their profile page) - I hope that's ok. 

I would suggest to keep the settings as the last tab, as it is more common in apps to have this (even here on Github the settings is the last item in the menu dropdown, for example). It makes sense because settings are not directly related to the content of the app, so it's a secondary thing.

> why is there a user name in addition to the name?

My idea was to have the real person's name + the user's name.

Here's the original mockup:

profile__overview

jh-RLI:
We can deactivate the delete functionality for now and develop it in its own branch. It would be good to add the tests first and then merge this PR as i want to use the new profile page to complete the open peer review process implementation by adding collaboration functionality for the contributing user.

wingechr:
@bmlancien sorry,I should have checked with the design.

* I changed the order again so that `Tables` comes first
* I disabled the `delete account` in UI and API

grafik


But there are still problems with the edit form when I click on edit profile:

* It says "Create new user*
* the field `name` is the already existing, mandatory user name field. To me it feels weird that I should add an optional  `username` field. (Ususally, it would be an optional full name field, right?)
* email: should there be a validation mail if mail changes
* password: It should not say `No password set`


grafik

bmlancien:
@wingechr 

> sorry,I should have checked with the design.

No problem, I should have posted the design here before.

> It says "Create new user*

Yes, I would suggest "Update account profile" or "Update settings"

> the field name is the already existing, mandatory user name field. To me it feels weird that I should add an optional username field. (Ususally, it would be an optional full name field, right?)

I think so too. A mandatory profile name and an optional real name make more sense.

PR #1196, opened 2023-04-10
Update oeplatform documentation

### LINKED ISSUE #753: Update API documentation on readthedocs?
The documentation on readthedocs does not provide an indication on whether it is up to date and when it was last updated, the copyright states 2016 which may lead users to think it is outdated.
grafik 

Thus I think it would be useful to add such an indicator as the main OEP page links prominently and directly to the API documentation on readthedocs: https://oep-data-interface.readthedocs.io/en/latest/api/how_to.html

As a follow-up I wonder: 
* Does this API documentation - and possibly also the rest on readthedocs -  need an update? How does its API part relate to the API tutorials? Do we replicate content and effort? 

cc: @stap-m @wingechr

jh-RLI:
update information about year and project 

closes #753


wingechr (commented):
approved

PR #1198, opened 2023-04-11
update jsoneditor to support bs5

wingechr:
The latest version of the json-editor library supports bootstrap5, so I upgraded it.

grafik

There is no issue attached to this commit.

@jh-RLI : is there anything else I need to consider?


jh-RLI (approved):
Works on my end.

Pleas add the updated json-editor to the changelog.

PR #1231, opened 2023-05-11
Hotfix/fix mirror command 2

wingechr:
*  mirror command: only ask to delete if something to do
* delete one time management command for metadata upgrade


jh-RLI:
@wingechr i added some commits related to version and changelog. Is this already mergeable?

wingechr:
yea. if you are ok with the changes in mirror.py, you can merge it

Issue #1205, opened 2023-04-13
[OPR] Implement open peer review UI


jh-RLI:
We are currently developing a prototype for the open peer review. Most of it is finished. Within the next two weeks we want to implement as much of it as we can manage. 

We have started to implement the back-end functionality, but significant parts are still under development. We will start partially implementing the user interface for all the parts that are functionally ready as soon as they are ready.

bmlancien:
Here are some of the mockups:

![db__model-draft--suggest](https://user-images.githubusercontent.com/33999424/232023282-cf3f1b11-ed14-4e11-b8ac-a779b2fa69d7.png)

![db__model-draft--suggest (1)](https://user-images.githubusercontent.com/33999424/232023304-e8959153-d2e9-4b13-90b4-4581576a657c.png)

jh-RLI:
Just noticed this. Not very urgent :)
![image](https://github.com/OpenEnergyPlatform/oeplatform/assets/38939526/9a931c1b-6e0d-412e-bca2-a97197cee56c)

jh-RLI:
This is now basically done. There are some parts missing, like the progress bar. They will be implemented in later iterations.

PR #1256, opened 2023-06-08
Fix/contributor value break line

bmlancien:
@jh-RLI I quickly fixed the last issue by forcing the line to break (#1205)


jh-RLI (approved):
Works for me.

PR #1355, opened 2023-08-16
Code cleanup

wingechr:
## Type of change 

only formatting via pre-commit (black + isort) so that later changes will not stand out


jh-RLI (approved):
Thanks for handling this.

PR #1482, opened 2023-11-29
Documentation: Virtual environments are created with the python executable, not with pip.

pajot:
Small documentation update.

### Updated
Virtual environments are created with the python executable, not with pip.


pajot:
The environment must be activated before installing requirements with pip.

jh-RLI (approved):
Good catch! Thanks for testing this.

Issue #1504, opened 2023-12-11
Implement a locking mechanism for editing the shared entities in bundles.


adelmemariani:
## Description of the issue

Certain entities within bundles are shared among multiple bundles. For example "Institutions" or "Contact person". As a result, multiple users can edit these entities simultaneously. This could potentially result in updates that are not consistent.  

## Ideas of solution

A locking mechanism that prevents simultaneous updates and only allows one update at a time based on a timeout. See below:
![Screenshot 2023-12-11 at 16 25 31](https://github.com/OpenEnergyPlatform/oeplatform/assets/7637364/4ead44a0-efc8-4060-bc92-2983411a0d1d)

## Workflow checklist
- [ ] I am aware of the workflow in [CONTRIBUTING.md](https://github.com/OpenEnergyPlatform/oeplatform/blob/develop/CONTRIBUTING.md)

PR #1609, opened 2024-03-02
adding locking mechanism for entities

h-spinde:
## Summary of the discussion

As mentioned in Issue #1504 , I added a locking mechanism so any entity currently being edited gets locked from being edited by other people.
An error message is currently not being displayed, as I edited only the backend, not the frontend. Furthermore, when editing an entity, for example an institution, the insititution label is always displayed as though it had been changed in the current tab - even when the edit is not saved (due to the entity being locked at the time of attempting to edit it). That the change has not been saved only become apparent if one leaves the current bundle, returns to scenario-bundles/main and then selects the bundle again:

after_editing_locked-entity-2

### Added
- Add a new functionality [(#)](https://github.com/OpenEnergyPlatform/oeplatform/pull/)

### Updated
- Update existing functionality [(#)](https://github.com/OpenEnergyPlatform/oeplatform/pull/

### PR-Assignee
- [ ] 📝 Update the [CHANGELOG.md](https://github.com/OpenEnergyPlatform/oeplatform/blob/develop/versions/changelogs/current.md)
- [ ] 📙 Update the documentation on [mkdocs](https://openenergyplatform.github.io/oeplatform-code/) 

### Reviewer

### DIFF
--- a/factsheet/models.py
+++ b/factsheet/models.py
@@ -29,3 +29,8 @@ class ScenarioBundleAccessControl(models.Model):
     @classmethod
     def load_by_uid(cls, uid):
         return ScenarioBundleAccessControl.objects.filter(bundle_id=uid).first()
+        
+class LockTable(models.Model):
+    resource_name = CharField(max_length=400)
+    timeout_time = DateTimeField(default=timezone.now)
+    transaction_id = CharField(max_length=400)
--- a/factsheet/views.py
+++ b/factsheet/views.py
@@ -23,6 +23,7 @@ from SPARQLWrapper import SPARQLWrapper, JSON
 import sys
 from owlready2 import get_ontology
 from pathlib import Path
+import datetime as dt
 
 from rest_framework.decorators import (
     api_view,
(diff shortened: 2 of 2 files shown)


jh-RLI:
Hey @h-spinde, great to see your contribution. You also have to push the django migration you created by running ´django manage.py makemigrations´ . There should be a file ` factsheet/migrations/0012_locktable.py` this needs to be availabe otherwiese the tests will not work.

h-spinde:
> Hey @h-spinde, great to see your contribution. You also have to push the django migration you created by running `django manage.py makemigrations` . There should be a file ` factsheet/migrations/0012_locktable.py` this needs to be availabe otherwiese the tests will not work.

Thank you! This was in fact enough to pass all the failed tests.

jh-RLI:
@adelmemariani could you start to review + test this? I will also test it soon. I think if we both agree this works as expected we can merge it. 

I think what is missing is the integration into the UI. The user should be able to see these "server" messages on the OEP-Website.

After that it would be grat to try it out on the TOEP.

jh-RLI:
I think we should also add a test for it (assertion), then we can be sure that future changes to the code will not affect this functionality. 

Also, I think we can move some of the functionality out of views.py to keep it cleaner as it is already very bloated. We can discuss where exactly the code goes. This is not immediately necessary.

jh-RLI:
I will change this to "draft Pull Request" until it is ready to be merged :)

PR #1612, opened 2024-03-13
fix open() with encoding, plus pre-commit code quality

wingechr:
Some environments do not have utf-8 as default character encoding.
Therefore, all `open("r")` statements need to specify utf-8 as  encoding, otherwise an error occurs (just happened on a test server)

I also had to add more changes because of the pre-commit code quality requirements, otherwise I could not commit.
@adelmemariani , @jh-RLI please use local pre-commit hooks so that changed filesalways follow the code guidelines.

Cheers,
C

### DIFF
--- a/api/references.py
+++ b/api/references.py
@@ -213,7 +213,7 @@ def read_bibtexfile(file_name):
     metadata.create_all(bind=engine)
     sess = Session(bind=engine)
 
-    with open(file_name) as bibtex_file:
+    with open(file_name, encoding="utf-8") as bibtex_file:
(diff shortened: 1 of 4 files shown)


jh-RLI:
Good catch!

I agree. We have started using it but it seems that not everyone has found it easy to use. I will demand it from now on. Could also introduce another ci action

jh-RLI (approved):
Nice! Two minor comments for future development:
1. we once agreed on branch names without / only - should be used 
(My bad the contributing.md is not up to date, will update it now)
2. Please always add a entry to the changelog `versions/changelogs/`

Issue #1394, opened 2023-10-11
Code cleanup: Restructure settings related code


jh-RLI:
We found some settings in the `factsheet/views.py` app that should be in the settings.py or securitysettings.py. We should also check for other code that might should be moved to a separate app config file or also to the global settings file.

PR #1643, opened 2024-03-27
Move endpoint variable to settings

h-spinde:
## Summary of the discussion

As part of solving issue [1394](https://github.com/OpenEnergyPlatform/oeplatform/issues/1394), the variables query_endpoint and update_endpoint should be moved from views.py to settings.py

### PR-Assignee
- [ ] 📝 Update the [CHANGELOG.md](https://github.com/OpenEnergyPlatform/oeplatform/blob/develop/versions/changelogs/current.md)
- [ ] 📙 Update the documentation on [mkdocs](https://openenergyplatform.github.io/oeplatform-code/) 

### Reviewer

### DIFF
--- a/factsheet/views.py
+++ b/factsheet/views.py
@@ -17,7 +17,7 @@ from rdflib.plugins.stores import sparqlstore
 from rdflib.namespace import XSD, Namespace
 from rdflib.graph import DATASET_DEFAULT_GRAPH_ID as default
 import os
-from oeplatform.settings import ONTOLOGY_ROOT, RDF_DATABASES, OPEN_ENERGY_ONTOLOGY_NAME
+from oeplatform.settings import ONTOLOGY_ROOT, RDF_DATABASES, OPEN_ENERGY_ONTOLOGY_NAME, query_endpoint, update_endpoint
 
 from SPARQLWrapper import SPARQLWrapper, JSON
 import sys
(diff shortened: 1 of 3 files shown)


jh-RLI:
That's great :) I would also like to draw your attention to some efforts I have made to make the code more modular:

I created a subfolder factsheet/oekg where this file is located https://github.com/OpenEnergyPlatform/oeplatform/blob/develop/factsheet/oekg/connection.py

I think we should use this for managing the connections. There you will see the same endpoints, these should also be imported from the settings. Then in the factsheet/views.py I suggest to import the factsheet/oekg/connection.py and use this for the oekg connection. In addition, we can then also remove this section https://github.com/OpenEnergyPlatform/oeplatform/blob/031ffe29b5597656e0f4634c025f1d67df218fe4/factsheet/views.py#L91-L108 as I have moved it to https://github.com/OpenEnergyPlatform/oeplatform/blob/develop/factsheet/oekg/namespaces.py. 

What do you think @adelmemariani ?

adelmemariani:
> What do you think @adelmemariani ?

If you tested it and everything is working it is ok. But, shouldn't the 'bind_all_namespaces' return a graph?

jh-RLI:
Further tests would certainly also be part of this branch, but in general everything has worked so far. I'm happy to help. 

You're right about your comment, it's been a while, I think there was another problem. 

It would be great if we could slowly make the code more modular starting with this branch.

jh-RLI:
As we are now using a different approach, this PR is outdated. Nevertheless, it was good that there was just another developer who suggested another working solution.

See: https://github.com/OpenEnergyPlatform/oeplatform/pull/1656

PR #1656, opened 2024-04-09
Changes for upgrade to python 310

wingechr:
- requirements updated for python 3.10
 - read knowledge graph endpoints from securitysettings.py


wingechr:
btw tests for 3.6/3.9 will stay pending here because they don't exist anymore, so you have to ignore that

PR #1748, opened 2024-06-25
Fix metadata order when downloading from data view

### LINKED ISSUE #1696: Inconsistens Metadata returns
## Description of the issue

The OEMetabuilder and the table download functions return metadata in a different form. 

## Steps to Reproduce

1. Create a table through the upload wizard and look at the JSON there
2. download the table metadata in the table view

## Ideas of solution

The two sources use different functions to return/create metadata. Either use the same function or if not feasible, adapt one function to match the output of the other.

Also see https://github.com/OpenEnergyPlatform/omi/issues/99

henhuy:
## Summary of the discussion

Fix inconsistent metadata structure

### Changes

- Metadata order when downloading from data view

### Automation

Closes #1696 

### PR-Assignee

### Reviewer

### DIFF
--- a/dataedit/helper.py
+++ b/dataedit/helper.py
@@ -4,6 +4,7 @@ make the codebase more modular.
 """
 
 from dataedit.models import Table
+from dataedit.metadata import TEMPLATE_V1_5
 
 ##############################################
 #          Table view related                #
@@ -53,6 +54,15 @@ def get_readable_table_name(table_obj: Table) -> str:
         raise e
     return label
 
+
+##############################################
+#            Metadata related                #
+##############################################
+
+def order_metadata(metadata: dict) -> dict:
+    """Brings metadata into consistent order based on TEMPLATE_V1_5"""
+    ordered_metadata = {key: metadata.get(key) for key in TEMPLATE_V1_5}
+    return ordered_metadata
 
 ##############################################
 #       Open Peer Review related             #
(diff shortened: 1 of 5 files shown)


versions/changelogs/current.md
```diff
@@ -1,6 +1,7 @@
 # Changes to the oeplatform code
 
 ## Changes
+- fixed metadata order when downloading metadata from data view
```
jh-RLI, line 4:
```suggestion
- fixed metadata order when downloading metadata from data view  [(#1748)](https://github.com/OpenEnergyPlatform/oeplatform/pull/1748)
```

dataedit/helper.py
```diff
@@ -54,6 +55,15 @@ def get_readable_table_name(table_obj: Table) -> str:
     return label
 
 
+##############################################
+#            Metadata related                #
+##############################################
+
+def order_metadata(metadata: dict) -> dict:
```
jh-RLI, line 62: 👍 If you agree that this is good enout. There is still a slight difference because this only sorts the "top" level buttons. I think it's fine.

jh-RLI (commented):
We have agreed that branch names should only contain "-" and no "/" or "_" :) but don't change this here, just as an information for future branch names.

dataedit/helper.py
henhuy: could adapt this also, so that key order is preserved recursively

jh-RLI: Right, that could be a quick solution. Then I would vote for implementing it : in favour of everything being as similar as possible.

PR #1870, opened 2024-09-24
Release from Master back to develop

jh-RLI:


jh-RLI:
Hmpf, the oeo 2.4.0 is no longer the latest release. This is why the CI fails as the OEOX owl file includes a link to the oeo, it must be updated to the latest release.

PR #1878, opened 2024-10-01
Updated oeo in docker 2.5

ebroda:
## Summary of the discussion
Update oeo in Docker to latest version

### Changes

- Updated oeo in docker image to version 2.5 [(#1878)](https://github.com/OpenEnergyPlatform/oeplatform/pull/1878)

### PR-Assignee

### Reviewer

- [x] 🐙 Follow the [Reviewer Guidelines](https://github.com/rl-institut/super-repo/blob/develop/CONTRIBUTING.md#40-let-someone-else-review-your-pr)
- [x] 🐙 Provided feedback and show sufficient appreciation for the work done


jh-RLI (approved):
Thank you very much! ... actually I thought for sure I've already made this update .... has been a little stressful the last few weeks. I would like to change this code so that it doesn't have to be changed every time an OEO release is available. So far I haven't come up with a good solution.

Changing the md file name to README is also good 👍

ebroda:
So we had it previously that the image is just pulling the latest release: https://github.com/OpenEnergyPlatform/oeplatform/commit/6e6e106bcbd96c533f006763711e374a369df36b . GitHub has a special URL for pulling the last one.

This lead to the problem with the 2.4.0 release, as that had a different structure in the archive than the release before leading to some problems and some "unwanted" directories. So if the OEO could establish a publishing process which ensures (basically) the same structure in the releases, probably we can go back to that and than have no need to update the line every time. Maybe some shell script doing the necessary steps and packaging of the files.

jh-RLI:
That's right! We already had it. 
As for the Oeo structure: I discussed this with them and they reverted the structure back to what was there before v2.3.0 (I think that was the point at which it changed) ... now I'm not 100% sure if they can keep it up as some of the developers can't continue working on the OEO as part of their job. We'll see on the next release, I'll try to take a look at the files before the release.

stap-m:
@ColinHDev can we include this into the release workflow?

ColinHDev:
> @ColinHDev can we include this into the release workflow?

What specifically? Check whether the `build-files.zip` has the right structure? Checking it automatically or mentioning it in the `How to release a new ontology` tutorial in the wiki?

PR #1914, opened 2024-12-02
Base templates: Apply DRY concept on base-templates

ebroda:
## Summary of the discussion
There are five base templates which are quite similar.

* base.html
* base-full.html
* base-profile.html
* base-sidebar.html
* base-wide-react.html

This Pull Request apply the DRY (Don't Repeat Yourself) concept, at least partly, on codes which have been identical in these files. Concrete these are the footer and navigation.

Additionally, links to github.com/openego/oeplatform have been updated to github.com/OpenEnergyPlatform/oeplatform.

### Changes

- Extract header/footer template [(#1914)](https://github.com/OpenEnergyPlatform/oeplatform/pull/1914)

### PR-Assignee

### Reviewer

- [x] 🐙 Follow the [Reviewer Guidelines](https://github.com/rl-institut/super-repo/blob/develop/CONTRIBUTING.md#40-let-someone-else-review-your-pr)
- [x] 🐙 Provided feedback and show sufficient appreciation for the work done


ebroda:
I think there's still some room for improvement / "Zusammenführung/Simplifizierung", as these files are still quite similar, mostly only differ by some classes which are somewhere set or not. But at least a start.

Regarding Auth: I build a variant for LDAP. If you want, I can also share that even though its a more specific case.

jh-RLI:
I agree ... there will be a funded project sometime next year just aimed at bug fixes/improvements. After that, I hope all these downside are finally fixed and the whole project is up to date using good practices. But I should take a look and just throw everything out of this “Wildwuchs” ;)

That's interesting! We will be using https://www.scc.kit.edu/dienste/regapp.php as there is a [hosted service](https://regapp.nfdi-aai.de/) we can use. It is also a proxy for LDAP services. The documentation is a bit opaque. I'm not an expert in this area, but I think this covers our needs for now. LDAP could be useful if we want tighter integration with other services we maintain, right? Federated Login seems to be sufficient for now.

PR #2005, opened 2025-05-14
docker compose: dont expose ports outside of internal network (except…

wingechr:
It seems you don't need to expose ports in docker compose, unless you explicitly need them outside the docker compose network (like debugpy and, obviously, web 8000). 

With these changes. there is no more conflict with my local postgres and its still working.

@jh-RLI what do you think?


jh-RLI:
Seems great in general, makes it cleaner. I wasn't thinking about that, as i just assumed i have to do this so the browser can access :)

PR #2365, opened 2026-07-03
Fix array fields dropping entries beyond the 9th on factsheet submit

### LINKED ISSUE #2328: Model Factsheet - General: More than 9 authors are dropped
## Description of the issue

When you add more than 9 `Authors` to a `Model` `Factsheet`, all authors exceeding 9 are dropped on submit.
They do not appear in the updated factsheet, and when re-editing the factsheet the previously entered fields (after the ninth one) are missing.

We have about 15 authors at the moment four our model and would like to name them all - if possible.

## Steps to Reproduce

1. Enter more than 9 authors in a model factsheet
2. submit
3. the 10th and forthfollowing authors are removed

## Ideas of solution

Check hard coded author limits or check for single-digit author identifiers.

If an author limit is intended (but why should it?): State that limit in the help text, do not allow to add further fields that are dropped.

## Context and Environment

- Version used: version live on https://openenergyplatform.org/ on May 27th 2026
- Operating system: Windows 11
- Environment setup and (python) version: n/a

## Workflow checklist

ajith1994:
## Summary of the discussion

Issue #2328: adding more than 9 entries to an array field on a Model/Framework
factsheet (e.g. Author(s)) causes everything from the 10th entry onward to be
silently dropped on submit, and not restored when re-editing.

Root cause is in `processPost()` (`modelview/helper.py`): the regex
`^{field}_\d$` matches only a single trailing digit, so field names like
`authors_10`, `authors_11` never match and are discarded before saving.
`parts.sort()` also did a lexicographic (string) sort rather than numeric,
which would have scrambled entry order once the truncation was fixed.

Fix: changed `\d$` to `\d+$`, and sort `parts` numerically by the trailing
index instead of lexicographically. Manually verified by submitting a Model
factsheet with 12 authors and confirming all 12 persist, in the correct
order, in `modelview_basicfactsheet.authors`.

### Bugs

- Fix Model/Framework factsheets silently dropping the 10th and later
  entries of array fields (e.g. Author(s)) on submit, caused by a regex
  that only matched single-digit field-name suffixes.
  [(#2365)](https://github.com/OpenEnergyPlatform/oeplatform/pull/2365)

### Automation

Closes #2328

### PR-Assignee

- [x] 📝 Update the


ajith1994:
Hi @jh-RLI,

Could you take a look when you get a chance? This fixes #2328 (Model/Framework
factsheets silently dropping array field entries beyond the 9th on submit).

This is my first contribution to the project. I will be happy to make any changes you'd like to see. And, please let me know if I have missed out anything.

Thank you!

jh-RLI:
@ajith1994 Thanks a lot for the contribution. I’ll take a closer look at it next week; at first glance, it seems like you fixed it.

jh-RLI (approved):
Thanks for the fix. I approve and will merge the PR. I Plan the next release to production within the Next 2 weeks.

PR #2398, opened 2026-08-03
Improve OpenAPI documentation grouping

tomi-rli:
## Summary of the discussion

Describe the findings of the discussion in the issue or meeting.

### Features

- Add OpenAPI schema, Swagger UI .

### Changes

- Group API endpoints into clearer functional categories
- Improve the OpenAPI documentation of dataset endpoints with summaries, parameters, responses and request examples.

### Bugs

- Fixed a problem with
  [(#)](https://github.com/OpenEnergyPlatform/oeplatform/pull/)

### Removed

- Remove a broken link
  [(#)](https://github.com/OpenEnergyPlatform/oeplatform/pull/)

### Documentation updates

- Updated documentation for
  [(#)](https://github.com/OpenEnergyPlatform/oeplatform/pull/)

### Automation

Closes #

### PR-Assignee

- [x] 📝 Update the

### Reviewer

- [ ] 🐙 Follow the


api/urls.py
```diff
@@ -15,6 +15,11 @@
 """  # noqa: 501
 
 from django.urls import include, path, re_path
+from drf_spectacular.views import (
+    SpectacularAPIView,
+    SpectacularRedocView,
+    SpectacularSwaggerView,
```
jh-RLI, line 21: We should only offer one of these, i think.

jh-RLI (changes requested):

api/urls.py
tomi-rli: Yes, i agree.  I think the important one is the SpectacularSwaggerView. The Redoc view is just an alternative.

jh-RLI: Sounds good. Lets use SpectacularSwaggerView

tomi-rli: i just delete the redoc view.

jh-RLI (commented):
## Review — verified by generating and validating the schema

I built the merged state, generated the schema with `manage.py spectacular`, validated it, and ran the tests. Three things needed fixing before this could land; I've pushed them as `3eb31ca` along with a merge of `develop`, so the branch is no longer conflicting.

### The headline result

**65 paths generate, against the 47 in the hand-written `docs/oeplatform-code/web-api/oedb-rest-api/schema.json`** — and every OEKG scenario-bundle endpoint appears with **no annotations at all**, docstrings picked up as descriptions. That is the whole case for generating rather than hand-writing, and the hand-written file makes it concrete: it has already lost the Dataset CRUD (#2071), Bulk Upload (v1.9.0) and everything OEKG.

### What I fixed (3eb31ca)

**1. The generated document was not valid OpenAPI.** `responses={}` on the dataset create annotation emitted a literal empty Responses Object, which the spec requires to be non-empty, so the validator rejected the entire schema on that one operation:

```
SchemaValidationError: {} should be non-empty
On instance['paths']['/api/v0/datasets/']['post']['responses']
```

It was also the least accurate annotation here: the view returns **201** with `{"id", "metadata"}`, and **without** the annotation the generator would have inferred something valid on its own — so it made that endpoint worse than leaving it alone. Now documents the 201 it really returns plus the 400 a rejected payload gets. `--validate` passes.

**2. Three of the four dataset field descriptions were in German**, beside a fourth in English (`"Anzeigename des Datensatzes"`, `"Kurze Beschreibung des Datensatzes"`, `"Optional: Persistenter Identifier oder URL für den Datensatz"`). They're read by anyone browsing the schema or the DRF browsable API, so they've followed the rest of the API into English.

**3. No changelog entry**, although this adds a user-visible page. Added under Features.

**Also merged `develop`** — the conflict was in `oeplatform/settings.py`, where `develop` added `DEFAULT_THROTTLE_RATES` for the OEKG bundle API and this branch adds `DEFAULT_SCHEMA_CLASS`. Both belong; resolved as a union.

### Verification

- `manage.py spectacular --validate` — passes.
- `manage.py test api oekg` — **246 tests, all green** (31 skipped: no local Fuseki).
- `pre-commit run` over the changed files — all hooks pass.

### Not fixed, worth knowing

- **The tag grouping is half-done, and it's this PR's headline.** Annotated views get curated tags (`Advanced`, `Advanced: Connection`, `Schema: Meta`); everything else gets a lowercase path-derived one (`datasets`, `tables`, `db`, `factsheet`, `oekg`, `schema`). Two conventions side by side — and `scenario-bundle` and `scenario-bundles` end up as two tags for one domain.
- **330 errors / 14 warnings**, nearly all pre-existing `unable to guess serializer` on plain `APIView`s — **not caused by this PR**. They're why the table and `advanced/` endpoints generate as bare `200: No response body`, and why the hand-written `schema.json` can't simply be deleted yet. Annotating those is separate work.
- `operationId "scenario_bundles_scenarios_retrieve" has collisions`, resolved as `…_retrieve2`. A poor name for a public schema, but it belongs to the OEKG endpoints, not to this PR.
- `/api/v0/schema/` overloads the word "schema", which means a Postgres schema everywhere else in this API (`/api/v0/schema/{schema}/tables/...`). Routing is correct — I checked both resolve — so this is a naming wart, not a bug.

Nothing here needs a change from @tomi-rli; leaving the merge decision to a maintainer.

jh-RLI (approved):
I updates some descriptions and extended the swagger integration a bit. I will now merge the PR

PR #2441, opened 2026-09-11
Record every write to a bundle, and read the record back

### LINKED ISSUE #2442: Every write to a scenario bundle leaves a record, and the record reads back
## Description of the issue

The audit trail for scenario bundles is silent almost everywhere. `OEKG_Modifications`
is written in exactly one place — `factsheet/views.py:1097`, the browser's update path.
Creating a bundle records nothing. Deleting one records nothing. `add_a_fact/`,
`add_entities/`, `update_an_entity/` and the `manage-datasets/` route record nothing.
And the REST API added in #2435 and #2438 now creates and patches bundles while
recording nothing at all, which makes the hole bigger with every slice.

Two consequences follow from that, and the second is the one that bites:

- **"Who created this bundle, and when" is unanswerable.** So is "who removed it".
- **"No entry" is ambiguous** between *this was created and never touched* and *this was
  changed by a path that does not record*. An audit trail with holes in it is not a
  weaker audit trail; it is one you cannot reason from at all.

The rows that do exist are also thin. A row carries `bundle_id`, two payload blobs, a
user and a timestamp — and nothing saying **which operation** it was or **which part** of
the bundle it touched. A timestamp cannot answer *which change produced the state I am
looking at*. Since #2438 every bundle carries a monotonic version, which makes that
answerable for the first time, and since #2435 sub-resources have identities — so the
missing columns are now fillable rather than merely missing.

Two more facts about the table, both verified in the working tree:

- **There is no index on `bundle_id`**, on this model or on `ScenarioBundleAccessControl`,
  and it is the only sensible query key for either.
- **The payloads are double-encoded**: `Graph.serialize()` returns a `str`, written into a
  `JSONField`, so every row holds a JSON string inside a JSON column. The existing React
  diff viewer depends on that.

This is the fifth of the twelve OEKG REST API slices, from the wayfinder map's tickets
note (*"Every write leaves a record, and the record reads back"*, derived from WF-08). It
is deliberately scheduled **early** — directly after the guarded patch and before
sub-resources — so that every later slice writes its own entries instead of being
retrofitted with them.

Separately filed and **not** part of this work: `get_oekg_modifications/` has no decorator
and serialises `OEKG_Modifications.objects.all()` in one response — every change to every
bundle, forever, to anonymous callers, including each acting user's raw FK pk.

### 1. Completeness: every write records

`POST`, `PATCH` and `DELETE` at every level write one entry. The rule is flat — **the
history is the log of writes** — with one stated exception that belongs to a later slice:

| Operation | Entry |
|---|---|
| `POST` bundle / sub-resource | entry + diff |
| `PATCH` bundle / sub-resource | entry + diff |
| `DELETE` sub-resource | entry + diff |
| `DELETE` bundle | event line only, no payload |

*Rejected: record only information-destroying writes (skip `POST`).* It halves the table's
growth and keeps the largest payloads out of it, but "who created this bundle and when"
stops being answerable and "no entry" stays ambiguous. A per-operation judgement about
what is worth recording is exactly what makes a trail patchy.

### 2. Entry content: a lossless triple diff plus structured columns

| Column | Value |
|---|---|
| `bundle_id` | the aggregate root, always present, **indexed** |
| `resource_type` | the OEO class touched |
| `resource_uuid` | the sub-resource, `NULL` for a bundle-level write |
| `verb` | `POST` / `PATCH` / `DELETE` |
| `user` | FK to `login.myuser`, as today |
| `timestamp` | as today |
| `version_before` / `version_after` | the monotonic version from #2438 |
| `removed` / `added` | the changed triples as **real** JSON, not double-encoded strings |

**The diff is free.** Validation already holds the post-state in memory and the
concurrency guard already holds the pre-state, so no extra graph read is needed — unlike
today's path, which re-reads the graph purely to compute it.

The diff stays **lossless and untranslated**: no field names, no serializer vocabulary, so
a later shape change never re-interprets an old row.

*Rejected: storing the request as sent.* Cheapest and trivially readable, but it does not
say what changed — a `PATCH` re-setting a field to its current value would be
indistinguishable from one that changed it. *Also rejected: storing a rendered field-level
summary alongside the triples.* More readable, and the hand-written serializers already
hold the predicate-to-field mapping — but it freezes one serializer version's reading into
a permanent row and can drift from the triples it describes. **Rendering is a read-side
concern**, computed from the triples at read time.

### 3. Visibility: public, per bundle, paginated

```
GET /api/v0/scenario-bundles/<uid>/history/
  200  public, paginated, actor as a username
```

**The reasoning that makes bundle reads public does not transfer here, and that was
checked rather than assumed.** That decision rests on "the SPARQL endpoint already serves
the same data" — but the history lives in Postgres and is in no graph. What it exposes is
different in kind, too: not bundle content but *who edited what, when*.

The distinction that carries it: **a per-bundle log is a record about a bundle; a global
one is a profile of a person's activity across the platform.** So this endpoint is scoped
to one bundle, paginated with a ceiling and no unbounded mode, and names the actor by
username rather than by the raw pk the existing global dump hands out.

*Rejected: public entries with an owner-only actor field.* The most defensible split under
data-protection scrutiny, but two representations of one resource to build and test, and
attribution is arguably the most useful field on a scientific record. *Also rejected:
authenticated, owner and admin only* — simplest privacy story, but a visible reduction
from today's behaviour that would take the existing changelog page dark for anonymous
visitors.

### 4. The seam: extend `OEKG_Modifications`

One migration on the existing model — not a new model, and **no rewrite of existing data**.
Legacy rows keep their double-encoded strings and get `NULL` in the new columns, readable
as "pre-API era" without a risky data migration, and the existing viewer keeps working
because it still receives strings.

*Rejected: normalise the legacy rows too.* A single representation and a single read path
— but it rewrites production audit data in place, which is the one table whose value is
being an unaltered record, and it breaks the existing viewer. *Also rejected: a new model
with the old one frozen.* No migration risk, but two histories split by which client did
the writing, which is this issue's own completeness problem reappearing in a new form.

### 5. Two transactions: the graph commits first

The graph write is one atomic SPARQL request; the history is Postgres. They cannot be one
transaction, and after decision 1 *every* write has a Postgres side effect.

**Order: SPARQL commits first, then the history row.** If the history write fails, the
request still answers success, the failure goes out as one structured log line, and the
gap is named in the response body.

Reasoning: the data is the truth, the history is the note about it. Returning a `500` for
a write that *succeeded* provokes exactly the retry that creates duplicates. The price is
named rather than hidden — **a gap in the audit trail is possible and visible only in the
log and in that one response field.**

*Rejected: history first, then graph.* No audit gap, and the diff is available early — but
a crash between the steps leaves a row for a change that never happened, and **a phantom
entry is worse than a gap, because it looks like truth.** *Also rejected: an outbox with
`pending`/`committed` status.* Neither gap nor phantom, but three steps and a status
column, and it only pays off if somebody actually inspects the pending rows.

**One implementation consequence that is easy to miss:** the history insert needs a
savepoint. Without one a real rejection poisons the surrounding transaction and the next
query raises, producing precisely the `500` this ordering exists to avoid. A test that
mocks the failure cannot catch this — a mocked exception poisons nothing — so the test has
to provoke a real database error.

### Acceptance criteria

- [ ] The existing modifications model is **extended**, not replaced: structured columns
      (bundle, resource type, sub-resource identifier, verb, actor, time, version before
      and after), an index on the bundle identifier, and payload columns holding real
      structured data.
- [ ] Legacy rows keep their double-encoded strings and remain readable as "pre-API era".
      No data migration rewrites existing audit rows.
- [ ] Creates and patches each write **one** entry carrying the added and removed triples.
      The diff is taken from the pre- and post-state the request already holds — **no extra
      read**.
- [ ] The stored diff is lossless and untranslated, so a later shape change never
      re-interprets old rows.
- [ ] **The graph commits first.** A failed history write answers success and emits a
      structured log line. The audit gap is named in the response contract and the logs,
      not hidden.
- [ ] A per-bundle history endpoint returns entries **paginated**, publicly, with the actor
      as a username — never a raw internal identifier.
- [ ] Entries render changes in **field names**, computed at read time. The stored triples
      remain reachable through a query parameter.
- [ ] Tests cover: each verb produces an entry with the right resource and version pair;
      the rendered summary matches the stored triples; the history endpoint paginates; a
      history write failure still returns a successful write.

Every slice carries its own tests; there is no final testing pass. Tests run against a real
graph store in the compose network and **skip cleanly** when it is unreachable, never fail.

## Deployment

**This one needs `python manage.py migrate`** — Django migration `factsheet.0012`. It is
the first migration this API work has needed; #2435 and #2438 needed none, so the deploy
story changes here. Without it every bundle write fails at the point it tries to record
itself.

The migration is **purely additive, and literally so**: seven nullable columns and one
index. No `UPDATE`, no `DEFAULT` backfill, nothing altered — rows written before the API
keep everything they have. This is deliberately unlike `modelview.0066`, where `AddField`
on an `auto_now_add` column silently stamped every existing row with the migration's own
timestamp.

Two things worth knowing before running it:

- The `CREATE INDEX` is **not** `CONCURRENT`, so it takes a write lock on the audit table
  for as long as it runs. That table holds one row per scenario-bundle edit made through
  the browser since 2023, so it is short — but it is a lock on the audit table, not
  nothing.
- The index is declared as a plain btree rather than `db_index=True`, because on a
  `CharField` that shortcut also builds a second `varchar_pattern_ops` index for the `LIKE`
  queries nothing here makes. `bundle_id` had no index at all before this.

**No `alembic upgrade head`, no new requirements, no configuration change.** The standing
requirement from #2428 is unchanged: `python manage.py fetch_oekg_shapes` must have run,
or the bundle endpoints answer `503`.

**One visible change to an existing page.** The scenario-bundle changelog under Factsheets
reads every row of this table, so it will list API writes too — with the columns it
understands and an empty diff, because the API records what changed in the newer columns
that page does not read.

**Rollback** is a code rollback plus leaving the migration in place; the new columns are
nullable and nothing reads them once the code is gone.

**Verifying the deploy** — reads are public, so this needs no token:

```
curl -s https://openenergyplatform.org/api/v0/scenario-bundles/<uid>/history/
```

A paginated body with `count`, `next` and `results` means it is live. An empty `results`
on a bundle the browser wrote is correct rather than a fault: nothing recorded its earlier
edits in the new form.

## Workflow checklist

Zwei Dinge dazu:

jh-RLI:
## Summary of the discussion

Fifth of the twelve slices of the OEKG REST API, from the wayfinder map's tickets note
(slice *"Every write leaves a record, and the record reads back"*, derived from WF-08).
It lands deliberately early — directly after the guarded patch — so that every later
slice writes its own entries instead of being retrofitted with them.

The audit trail stops being silent. Today only the browser's update path writes an entry;
creates and every API write leave nothing behind. The rule here is flat — **the history is
the log of writes** — because a per-operation judgement about what is worth recording is
what makes a trail patchy, and "no entry" would then be ambiguous between *created* and
*never touched*.

**The existing model is extended rather than replaced.** Two histories split by which
client did the writing is this table's own completeness problem in a new form, so the API
writes here. That leaves two generations of row and they are deliberately not reconciled:
pre-API rows keep their JSON-LD strings inside a JSON column, untouched by any data
migration, because rewriting them in place would edit the one table whose value is being an
unaltered record — and would break the diff viewer that reads them. `verb IS NULL` is what
tells the generations apart. New rows fill the columns that make the table answerable:
which operation, which resource, which version pair. A timestamp cannot say *which change
produced a given state*; `version_before` and `version_after` can.

**The stored diff is lossless and untranslated** — no field names, no serializer
vocabulary — so a later shape change never re-interprets an old row. Field names are
computed instead at read time, from the same field table that builds the triples, so a
rendering cannot drift from what it describes. A triple that names no field is reported
under a null field rather than dropped: the type and label a minted contact brings with it
are not bundle fields, and dropping them would make a summary look complete when it is not.
Where a predicate is shared — frameworks and models both hang off has-part — the type in
the diff decides, and an unlinked part carries no type, so the field is reported as unknown
rather than guessed at. Every rendered change also names its predicate, so an entry that
could not be attributed still says what it was about.

**The graph commits first and a failed history write does not fail the request.** The two
stores cannot share a transaction; the data is the truth and the history is the note about
it, so answering `500` for a write that succeeded would provoke exactly the retry that
creates duplicates. The gap is named rather than hidden — `_meta.history_recorded` and one
structured log line — because a phantom entry would be worse: it looks like truth.

Two things about that which are worth the reviewer's attention, because both were found by
reviewing rather than by writing:

- **The insert takes a savepoint, and that is not decoration.** Without it a real rejection
  poisons the surrounding transaction and the next query raises, producing precisely the
  `500` this ordering exists to avoid. The test for it uses a real over-long value rather
  than a mock, because a mocked exception cannot poison anything — the first version of that
  test passed over the hole.
- **`record_write` says it never raises, and at first it did not hold.** Only
  `DatabaseError` was caught while serialising the diff ran inside the same `try`, so an
  rdflib failure would have escaped and turned a committed graph write into a `500`. The net
  now covers every exception, which is load-bearing rather than defensive.

**Reads are public but per bundle and paginated with a ceiling**, and name the actor by
username. WF-07's reasoning for public reads does not transfer: it rests on the SPARQL
endpoint already serving the same data, and this is in no graph. A per-bundle log is a
record about a bundle; a global one is a profile of a person's activity — which is what the
existing global dump is, and it is filed separately.

Two smaller things found while wiring it up. **API rows write `""` into the legacy payload
columns rather than `NULL`**, because the existing diff viewer hands both straight to a
component that splits them and renders every row in one pass, so one null would blank the
whole page. And **`bundle_exists` asks with an `ASK` and lets a store outage raise**: "there
is no such bundle" and "I could not find out" are different answers, and only one of them is
a `404`.

### Features

- Every write to a scenario bundle through the REST API now leaves a record,
  and `GET /api/v0/scenario-bundles/<uid>/history/` reads it back. Previously
  only the browser's edit path recorded anything, so creates and API writes were
  silent. An entry says which operation it was, who made it, when, which version
  it produced, and which triples changed; the reader renders those as field
  names and returns the raw triples on `?expand=triples`. The history is public
  but per bundle and paginated, and names the actor by username rather than by
  internal id. Entries written before this release are kept exactly as they are
  and read as coming from before the API
  [(#2441)](https://github.com/OpenEnergyPlatform/oeplatform/pull/2441)

### Changes

- The scenario-bundle changelog page under Factsheets now also lists changes
  made through the REST API. Those rows show who changed what and when, but no
  side-by-side diff: the API records what changed in a newer form that this page
  does not read yet
  [(#2441)](https://github.com/OpenEnergyPlatform/oeplatform/pull/2441)

### Bugs

None as a separate entry. Nothing released was broken: both defects fixed here —
the un-savepointed insert and the too-narrow exception net — were introduced and
fixed inside this PR. Recording them under Bugs would tell a reader of the release
notes that something they were running had failed, which is not true.

### Removed

None.

### Documentation updates

None, for the same reason as #2438: the API's own documentation is a separate,
currently blocked strand of the map. WF-13 (where the API description comes from)
gates WF-14 (developer documentation), `drf-spectacular` is already wired on an
unmerged branch that ticket has to reconcile with, and the existing
`docs/oeplatform-code/web-api/oekg-api/oekg.yaml` describes the SPARQL passthrough
rather than the bundle CRUD. The reasoning lives in the module docstrings meanwhile
— `oekg/history.py` states why the diff is stored untranslated and why the graph
commits first.

**This is now the second slice to leave documentation open, and the gap is the map's
actual frontier.** WF-13 has been the next unblocked ticket since 2026-09-08 and
nothing has been done on it, while each slice adds endpoints that will have to be
described afterwards.

### Automation

Closes #2442

### PR-Assignee

- [x] 📝 Update the
      [mkdocs](https://openenergyplatform.github.io/oeplatform/) — deliberately not
      done, see *Documentation updates* above

### Reviewer

- [ ] 🐙 Follow the

### DIFF
--- a/api/urls.py
+++ b/api/urls.py
@@ -73,6 +73,7 @@ from api.views import (
     usrprop_api_view,
 )
 from oekg.api_views import ScenarioBundleAPIView, ScenarioBundleCollectionAPIView
+from oekg.history_views import ScenarioBundleHistoryAPIView
 
 app_name = "api"
 
@@ -311,6 +312,11 @@ urlpatterns_v0 = [
         "scenario-bundles/<uid>/",
         ScenarioBundleAPIView.as_view(),
         name="scenario-bundle",
+    ),
+    path(
+        "scenario-bundles/<uid>/history/",
+        ScenarioBundleHistoryAPIView.as_view(),
+        name="scenario-bundle-history",
     ),
     path(
         "datasets/",
--- a/factsheet/migrations/0012_oekg_modifications_api_history.py
+++ b/factsheet/migrations/0012_oekg_modifications_api_history.py
@@ -1,21 +1,26 @@
 """Room in the modifications table for what the REST API records.
 
-Purely additive: every new column is nullable and nothing backfills. Rows
-written before the API keep their values untouched -- `verb` stays NULL on
-them, which is what lets a reader tell the two generations apart without a
-migration that rewrites the one table whose value is being an unaltered record.
+**Purely additive, and literally so**: seven nullable columns and one index.
+Nothing backfills, nothing is altered, and no existing value is read or
+rewritten. Rows written before the API keep everything they have -- `verb`
+stays NULL on them, which is what lets a reader tell the two generations apart
+without touching the one table whose value is being an unaltered record.
 
-`old_state` and `new_state` become nullable for the same reason from the other
-side: API rows carry their diff in `removed` / `added` as real structured JSON,
-so they leave the double-encoded legacy columns empty rather than writing a
-second representation into them and breaking the existing diff viewer.
+The index is the only non-column change. `bundle_id` is the sole query key this
+table has and it had no index at all. Declared as a plain btree rather than
+`db_index=True`, because on a CharField that shortcut also builds a second
+varchar_pattern_ops index for LIKE queries nothing here makes.
 
-The index on `bundle_id` is the only sensible query key and there was none.
+Deploy note: the CREATE INDEX is not CONCURRENT, so it takes a write lock on
+the table for as long as it runs. The table holds one row per scenario-bundle
+edit made through the browser since 2023, so that is short -- but it is a lock
+on the audit table, not nothing.
 
 SPDX-FileCopyrightText: 2026 Jonas Huber <https://github.com/jh-RLI> © Reiner Lemoine Institut
 SPDX-License-Identifier: AGPL-3.0-or-later
 """  # noqa: 501
 
+from django.conf import settings
 from django.db import migrations, models
 
 
@@ -23,6 +28,7 @@ class Migration(migrations.Migration):
 
     dependencies = [
         ("factsheet", "0011_scenariobundleaccesscontrol"),
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
     ]
 
     operations = [
@@ -61,19 +67,8 @@ class Migration(migrations.Migration):
             name="version_before",
             field=models.IntegerField(blank=True, null=True),
         ),
-        migrations.AlterField(
+        migrations.AddIndex(
             model_name="oekg_modifications",
-            name="bundle_id",
-            field=models.CharField(db_index=True, default="none", max_length=400),
-        ),
-        migrations.AlterField(
-            model_name="oekg_modifications",
-            name="new_state",
-            field=models.JSONField(blank=True, null=True),
-        ),
-        migrations.AlterField(
-            model_name="oekg_modifications",
-            name="old_state",
-            field=models.JSONField(blank=True, null=True),
+            index=models.Index(fields=["bundle_id"], name="oekg_mod_bundle_idx"),
         ),
     ]
--- a/factsheet/migrations/0012_oekg_modifications_api_history.py
+++ b/factsheet/migrations/0012_oekg_modifications_api_history.py
@@ -1,0 +1,79 @@
+"""Room in the modifications table for what the REST API records.
+
+Purely additive: every new column is nullable and nothing backfills. Rows
+written before the API keep their values untouched -- `verb` stays NULL on
+them, which is what lets a reader tell the two generations apart without a
+migration that rewrites the one table whose value is being an unaltered record.
+
+`old_state` and `new_state` become nullable for the same reason from the other
+side: API rows carry their diff in `removed` / `added` as real structured JSON,
+so they leave the double-encoded legacy columns empty rather than writing a
+second representation into them and breaking the existing diff viewer.
+
+The index on `bundle_id` is the only sensible query key and there was none.
+
+SPDX-FileCopyrightText: 2026 Jonas Huber <https://github.com/jh-RLI> © Reiner Lemoine Institut
+SPDX-License-Identifier: AGPL-3.0-or-later
+"""  # noqa: 501
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("factsheet", "0011_scenariobundleaccesscontrol"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="oekg_modifications",
+            name="added",
+            field=models.JSONField(blank=True, null=True),
+        ),
+        migrations.AddField(
+            model_name="oekg_modifications",
+            name="removed",
+            field=models.JSONField(blank=True, null=True),
+        ),
+        migrations.AddField(
+            model_name="oekg_modifications",
+            name="resource_type",
+            field=models.CharField(blank=True, max_length=400, null=True),
+        ),
+        migrations.AddField(
+            model_name="oekg_modifications",
+            name="resource_uuid",
+            field=models.CharField(blank=True, max_length=400, null=True),
+        ),
+        migrations.AddField(
+            model_name="oekg_modifications",
+            name="verb",
+            field=models.CharField(blank=True, max_length=10, null=True),
+        ),
+        migrations.AddField(
+            model_name="oekg_modifications",
+            name="version_after",
+            field=models.IntegerField(blank=True, null=True),
+        ),
+        migrations.AddField(
+            model_name="oekg_modifications",
+            name="version_before",
+            field=models.IntegerField(blank=True, null=True),
+        ),
+        migrations.AlterField(
+            model_name="oekg_modifications",
+            name="bundle_id",
+            field=models.CharField(db_index=True, default="none", max_length=400),
+        ),
+        migrations.AlterField(
+            model_name="oekg_modifications",
+            name="new_state",
+            field=models.JSONField(blank=True, null=True),
+        ),
+        migrations.AlterField(
+            model_name="oekg_modifications",
+            name="old_state",
+            field=models.JSONField(blank=True, null=True),
+        ),
+    ]
--- a/factsheet/models.py
+++ b/factsheet/models.py
@@ -59,9 +59,9 @@ class OEKG_Modifications(models.Model):
     row.
     """
 
-    bundle_id = CharField(max_length=400, default="none", db_index=True)
-    old_state = JSONField(null=True, blank=True)
-    new_state = JSONField(null=True, blank=True)
+    bundle_id = CharField(max_length=400, default="none")
+    old_state = JSONField()
+    new_state = JSONField()
     user = ForeignKey("login.myuser", on_delete=models.CASCADE, null=True)
     timestamp = DateTimeField(default=timezone.now)
 
@@ -73,6 +73,14 @@ class OEKG_Modifications(models.Model):
     version_after = IntegerField(null=True, blank=True)
     removed = JSONField(null=True, blank=True)
     added = JSONField(null=True, blank=True)
+
+    class Meta:
+        # A plain btree, declared here rather than as `db_index=True` on the
+        # field: on a CharField that shortcut also creates a second index with
+        # varchar_pattern_ops, for the LIKE queries nothing here makes. One
+        # index is the one this table needs -- it is queried by bundle and had
+        # none at all.
+        indexes = [models.Index(fields=["bundle_id"], name="oekg_mod_bundle_idx")]
 
     @property
     def era(self) -> str:
--- a/factsheet/models.py
+++ b/factsheet/models.py
@@ -9,16 +9,75 @@ SPDX-License-Identifier: AGPL-3.0-or-later
 """  # noqa: 501
 
 from django.db import models
-from django.db.models import CharField, DateTimeField, ForeignKey, JSONField
+from django.db.models import (
+    CharField,
+    DateTimeField,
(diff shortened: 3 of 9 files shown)


jh-RLI:
## Deployment

**This needs `python manage.py migrate`** — Django migration `factsheet.0012`. It is
the first migration this API work has needed; #2435 and #2438 needed none, so the
deploy story changes here. Without it every bundle write fails at the point it tries
to record itself.

The migration is **purely additive, and literally so**: seven nullable columns and one
index. Verified with `sqlmigrate` — no `UPDATE`, no `DEFAULT` backfill, and nothing
altered. Rows written before the API keep everything they have. This is deliberately
unlike `modelview.0066`, where `AddField` on an `auto_now_add` column silently stamped
every existing row.

Two things worth knowing before running it:

- The `CREATE INDEX` is **not** `CONCURRENT`, so it takes a write lock on the audit
  table for as long as it runs. That table holds one row per scenario-bundle edit made
  through the browser since 2023, so it is short — but it is a lock on the audit table,
  not nothing.
- The index is declared as a plain btree rather than `db_index=True`, because on a
  `CharField` that shortcut also builds a second `varchar_pattern_ops` index for `LIKE`
  queries nothing here makes. `bundle_id` had no index at all before this.

No `alembic upgrade head`, no new requirements, no configuration change. The standing
requirement from #2428 is unchanged: `python manage.py fetch_oekg_shapes` must have run,
or the bundle endpoints answer `503`.

**Verifying the deploy** — reads are public, so this needs no token:

```
curl -s https://openenergyplatform.org/api/v0/scenario-bundles/<uid>/history/
```

A paginated body with `count`, `next` and `results` means the slice is live. An empty
`results` on a bundle the browser wrote is correct rather than a fault: nothing recorded
its earlier edits in the new form.

## Reviewer's guide to the diff

Three commits: the slice, then the review's yield, then the changelog.

| File | What to look at |
|---|---|
| `factsheet/models.py` + `migrations/0012` | The seam. Two generations of row in one table, told apart by `verb IS NULL`, with the reasoning for not reconciling them in the model docstring. |
| `oekg/history.py` | `record_write` — the ordering, the savepoint, and the contract that it never raises. `changed_fields` — rendering computed at read time, and what it does with a triple it cannot attribute. |
| `oekg/history_views.py` | Public, per bundle, paginated with a ceiling. The docstring says why WF-07's public-reads argument does not transfer. |
| `oekg/api_support.py` | New: the throttles, the shared refusals and the existence check. Extracted when the second view module had to import five names from the first, and five more view modules are coming. |
| `oekg/api_views.py` | Where the two `record_write` calls sit — after the graph has committed, outside the `try` that could still turn an error into a response. |
| `oekg/tests/test_bundle_history.py` | 41 tests. `RealHistoryFailureTest` is the one to read: it uses a real over-long value, because a mocked exception cannot poison a transaction and so cannot test the savepoint. |

**Tests: 850 green** (807 before this branch), against a real Fuseki and the fetched shape
artifacts. Without a store or artifacts the graph tests skip with a stated reason.

PR #2469, opened 2026-09-15
Say whether a linked dataset is still there

### LINKED ISSUE #2468: Say whether a linked dataset is still there
## Description of the issue

A scenario in a bundle can say "I consumed table X" and "I produced dataset Y" (#2451), but
nothing behind either statement is enforced. The citation is a string in Fuseki; the target
is a row in Postgres; the two have different owners. So a reader of a bundle cannot tell a
live citation from a dead one, and the server — which could answer that in two indexed
queries — says nothing.

This is the eleventh of the twelve OEKG REST API slices, from the wayfinder map's tickets
note (*"Say whether a linked dataset is still there"*). It is independent of the other two
open strands and is collected, with them, by *"Replace the whole bundle in one call"*.

**Three ways the ground moves under a citation, and only one of them is a deletion:**

- A **table** can be deleted through the existing API (`TableAPIView.delete`,
  `api/views.py`). The graph keeps the dead link silently.
- An **OEP dataset** can be deleted too — and it can also become *unmaintainable* without
  being deleted, because `Dataset.creator` is `SET_NULL` (`dataedit/models.py:482-488`), so
  deleting a user leaves an ownerless dataset behind.
- A **dataset's membership** can change with no deletion anywhere: a table joins or leaves
  it. A bundle that cited the dataset then means something different than when it was
  written. That is the flip side of the coarser reference the resource model deliberately
  allowed.

Alongside availability there is **quality**: `PeerReview.table` (`dataedit/models.py:585`)
is a `CharField` holding the table *name* — the same key a dataset link already carries — so
whether the cited data has been through Open Peer Review is reachable in the same batched
lookup, with no new mechanism.

### 1. Never block, never collect, resolve on read

A dead link **stays** and the target's deletion is **never blocked**. Two reasons pointing
in opposite directions from the two rejected options:

- A scenario bundle is a **published research record**. "This scenario used table X" stays
  historically true after X is deleted. Garbage-collecting the link would falsify the
  record, and it would change a bundle without its owner doing anything.
- Blocking would give any user a **veto over other people's data** — anyone could make a
  stranger's table undeletable by citing it. That inverts the platform's permission model
  and would make table deletion depend on Fuseki being up.

### 2. Computed per read, batched by target kind

A read collects the referenced names and issues **one** query per target kind against the
same Postgres the API process already uses. Nothing is stored, nothing can go stale, and no
job runner is required — the line `Dataset.resource_entries()` already takes.

Rejected: *let the client discover the 404* (every client re-implements the same check, and
the representation withholds something the server knows cheaply). Also rejected: *a periodic
sweep stamping a date into the graph* — it would yield a genuinely useful **date**, but there
is no scheduler in this repo, and the sweep would write into other people's bundles, which
under the history and concurrency rules manufactures history entries and version bumps that
no user caused.

### 3. A dataset reference stays live, and the drift is made visible

A referenced OEP dataset whose membership changes **means what it means today**. A read
resolves it to the dataset's **current** members. The contract states the trade-off in one
line: `ref: table` is the reproducible reference, `ref: dataset` the current one.

Rejected: *snapshot the resolved table list at write time*. For a scientific record that is
arguably the right semantics, but `ex:DatasetShape` is `sh:closed` and the writable payload
is fixed at `{type, ref, name}` — a snapshot needs a **new predicate**, i.e. a change to the
shape, which belongs to the SHACL effort first.

### 4. The review indicator is three-valued

`finished`, `in_progress`, or **absent**. A boolean would collapse "never entered the
process" and "failed it" into one falsey value, and would read as a quality judgement on the
great majority of the platform's data, which predates the review process.

### 5. No hook in the table or dataset delete paths

`TableAPIView.delete` and `DatasetAPIView.delete` stay untouched; nothing in the table
subsystem learns about Fuseki. After decision 1 the consequence is visible where it matters
— in the bundle read.

Rejected: *warn without blocking* (an `ASK` in the delete path saying "deleted — 3 scenario
bundles referenced this"). It tells the human at the right moment, but it couples the table
subsystem to Fuseki and must be skipped silently when Fuseki is down — a warning nobody can
rely on.

### Acceptance criteria

- [ ] Every dataset link reports whether its target still exists, computed **per read** with
      one batched query per target kind. Nothing is stored and nothing can go stale.
- [ ] A dead link is **never** removed and **never** blocks its target's deletion.
- [ ] A dataset-kind reference resolves to the target's **current** member tables, and the
      documentation states the trade-off.
- [ ] A peer-review indicator is **three-valued**: finished, in progress, or absent.
- [ ] No hook is added to the table or dataset deletion paths.
- [ ] Existence does **not** imply maintained — a target whose creator record is gone still
      resolves.
- [ ] Tests cover: a deleted target reads as unresolvable and the link is still present; a
      dataset-kind link reflects a membership change; all three review states; the read
      issues a fixed number of relational queries regardless of how many links a bundle has.

Every slice carries its own tests; there is no final testing pass. Tests run against a real
graph store and **skip cleanly** when it is unreachable, never fail.

## Deployment

**No migration, no new dependency, no configuration change.** Like #2443, #2451 and #2456,
this slice changes what existing routes answer.

The standing requirement from #2428 is unchanged: `python manage.py fetch_oekg_shapes` must
have run, or the bundle endpoints answer `503`.

## Workflow checklist

jh-RLI:
Closes #2468.

Eleventh of the twelve OEKG REST API slices. A dataset link has always been a citation
with nothing behind it — the link is in Fuseki, its target is a row in Postgres, and the
two have different owners. This slice does not tighten that coupling; it makes the
consequence **visible**, which is the interim state the ticket deliberately settled on.

## Never block, never collect, resolve on read

A dead link stays, its target stays deletable, and every read says what the citation means
now. Both halves of that have a rejected opposite worth naming:

- **Collecting the link would falsify a published record.** "This scenario used table X"
  stays historically true after X is gone, and a collector would edit somebody's bundle
  without them touching it.
- **Blocking would invert the permission model.** Anyone could make a stranger's table
  undeletable by citing it, and table deletion would start depending on Fuseki being up.

So `_meta` on every link body gains two keys:

```json
{
  "type": "input", "ref": "table", "name": "abbb_emob",
  "_meta": {
    "target_iri": "https://openenergyplatform.org/database/tables/abbb_emob",
    "resolvable": true,
    "tables": [{"name": "abbb_emob", "peer_review": "finished"}]
  }
}
```

`tables` is the citation's meaning **today**: the one table for `ref: table`, the
catalogue entry's current members for `ref: dataset`, `[]` when the named target is gone.
A membership change shows up in the next read with no write to the bundle — that currency
is what `ref: dataset` is *for*, and the contract states the trade-off: a table reference
is the reproducible one, a dataset reference the current one.

Nothing is stored, so nothing can go stale. A link written against a table that did not
exist starts resolving the moment the table appears, and the bundle's version does not
move — a test pins exactly that, because the alternative (a periodic sweep stamping a date
into the graph) would write into other people's bundles and manufacture history entries
and version bumps no user caused.

**No hook in `TableAPIView.delete` or `DatasetAPIView.delete`.** The table subsystem still
knows nothing about Fuseki.

## Three things that read like bugs and are decisions

**1. The review indicator is three-valued.** `finished`, `in_progress`, or `null` — never
`false`. Open Peer Review postdates most of the platform's data, so a boolean would
present the great majority of links as having *failed* review. It sits on each entry in
`tables` rather than on the link, because `PeerReview.table` is keyed per table and there
is no such thing as a reviewed catalogue entry. A table with several reviews reads
`finished` if any of them finished: having been through review is not retracted by a later
round starting.

**2. An unroutable link reads `null`, not `false`.** The live graph holds databus URLs.
`ref` already reads back null for those, because the writable payload cannot express such
an address, and resolution says the same thing the same way — answering `false` would
claim the target had been deleted, which is a fabrication rather than an answer. (The
round-trip hole that pins is still slice 12's to decide; nothing here changes it.)

**3. `resolvable: true` does not mean maintained.** `Dataset.creator` is `SET_NULL`, so a
live target can be one nobody owns. Pinned by a test rather than left to be rediscovered.

## One live defect found while building it

Resolution first keyed on the link's `name` — which is its `rdfs:label`, while `ref` is
read out of its stored URL. For links this API wrote the two agree, because the URL is
built from the name. For links the older `manage-datasets/` route wrote they do not:
`oekg/sparqlModels.py` fills `dataset_label` from `name` and `dataset_url` from
`external_url`, two independent client-supplied values. A legacy link can therefore carry
a real table's address beside a human-readable title — and resolving by that title reported
a table that is plainly there as **deleted**, the exact fabrication the module refuses
everywhere else.

`reference_target` now reads both halves out of the URL. The payload's `name` still
reports the stored label, because that is what a write accepts. A URL reaching past a
target's own page — a table's permissions page — names nothing to resolve, because a page
*about* a table is not the table.

## Where it does not appear, and why

**Not in the bundle read.** A bundle read nests scenarios and study reports but no dataset
links (that is slice 7's shape), so resolution is on the three link endpoints only. WF-11's
"visible in the bundle read" predates the links having their own URLs; whether the bundle
read nests them is the read side's question, and that slice is in flight.

**And in `_meta`, not beside `ref`,** where WF-11 sketched it. WF-12 settled that the
payload's top level is exactly what a write accepts, so a client can send back what it read
without stripping — the property `replace` depends on. `target_iri` already landed there in
#2452 for the same reason.

## Cost

Three relational queries for a read, whatever the scenario cites: one per kind of target
named, and one for the reviews of every table those turned up. `ResolutionCostTest` pins
the absolute **3** as well as its independence from the number of links — two equal numbers
would also be the answer if resolution silently stopped happening.

Each target kind is looked up through an entry in `RESOLVERS` rather than a branch, keyed
by the same names `TARGETS` uses, and `ResolverCoverageTest` fails until a new kind gets
one. Silence is the dangerous failure here: a kind with no resolver would read as data that
had been deleted.

## Tests

**24 new** in `oekg/tests/test_dataset_link_resolution.py`; **1080 green** overall
(`manage.py test`, real Fuseki, real Postgres). `ExternalLinkMixin` was extracted from
#2452's `LinkWrittenElsewhereTest` so both modules build the same legacy link.

Covered: a deleted target reads unresolvable and the link is still there; a catalogue
membership change in either direction; an empty catalogue entry told apart from a deleted
one; all three review states plus the multi-round rule; a review outliving its table; an
ownerless entry; resolution by URL where the label differs; the query count; the unroutable
case.

## Deployment

**No migration, no new dependency, no configuration change.** The standing requirement from
#2428 is unchanged: `python manage.py fetch_oekg_shapes` must have run.

Documentation checkbox deliberately unticked — per the maintainer's call of 2026-09-15 the
OEKG API docs land in one pass once the code is there.


jh-RLI:
Filed #2473 out of this slice: dataset links are `BUNDLE_LOCAL_CLASSES` for deletion but are not in the bundle payload, so the replace slice's delete-by-omission would destroy every link in a bundle unless it is decided first. The round-trip hole `LinkWrittenElsewhereTest` pins is the second half of the same question and is collected there. Nothing in this PR changes either half — it only makes an unroutable link resolve to `null` rather than `false`.

jh-RLI:
Merged develop in (`3aab3a840`) after #2470 and #2472 landed. Both conflicts were the ones flagged when this branch opened, and both were additive:

- **`dataset_link_views.py`** — #2472 gave `_link_body`/`_link_bodies` an `expand` parameter and wrapped the body in `labelled(...)`; this branch gave them a resolution pass. Neither displaces the other (`labelled` fills `_meta.labels` when asked, resolution fills `_meta.resolvable`/`_meta.tables` always), so both sides were taken.
- **the changelog** — a plain concatenation.

Two things the merge could not check, checked by hand: `represented` now folds `write.gaps` into `_meta` (#2470), which lands before resolution and touches different keys; and `resolving()` reads the expansion set `OekgAPIView.initial` parses before dispatch, so a bad `expand` on a link `POST` still refuses before anything is written.

Added `ResolutionAndLabelExpansionTest`, because nothing asserted that the two read-only additions coexist — exactly the property a merge resolution can break silently. It also pins that resolution is **not** an expansion: a reader needs it to tell a live citation from a dead one, so it is never opt-in.

**1206 tests green** locally on the merged tree (real Fuseki, real Postgres).

PR #2489, opened 2026-09-21
Check a pooled connection before handing it to a request (#2488)

### LINKED ISSUE #2488: Advanced API serves dead pooled connections (pool_pre_ping absent) — real, but NOT the cause of the reported table-view 400s
## What happens

Opening a table's data view on production often answers the first
`POST /api/v0/advanced/search` with

```
400 Bad Request
{"reason": "Invalid request"}
```

Reloading two or three times makes it work; it comes back later.

## Cause, reproduced

`{"reason": "Invalid request"}` is not a message about the query. It is the
catch-all in `api/helper.py:238` — every exception that is not an `APIError`
becomes that body.

The exception is a dead pooled database connection. The OEDB engine
(`oedb/connection.py:34`) is created with

```python
create_engine(..., pool_size=0, pool_recycle=600, max_overflow=200)
```

`pool_recycle` discards a connection for being **old**, never for being
**dead**, and there is no `pool_pre_ping`. When Postgres or the network has
closed a pooled connection, the next checkout gets the corpse and psycopg2
raises. SQLAlchemy then invalidates it, so the retry opens a fresh one — which
is why a reload "fixes" it and why it returns later.

Reproduced end to end against a local instance: warm the pool with one search,
`pg_terminate_backend` the idle backend, search again.

```
WARM:            (200, '{"content": {...}, "data": [[0]], ...')
KILLED backends: 1
AFTER KILL [0]:  (400, '{"reason": "Invalid request"}')
AFTER KILL [1]:  (200, ...)
```

The swallowed exception is `server closed the connection unexpectedly`.

Adding `pool_pre_ping=True` and rerunning the same probe: the request after the
kill answers `200`.

## Why the table page is where you notice it

`dataedit/static/dataedit/backend.js:468` fires the count query and the select
query **in parallel** for every draw, so a table view takes two pool checkouts
at once and either one failing fails the whole draw.

## Why now

The fault is old; the way it presents is new.

- The pool settings are unchanged since `1a1bf52c2` (2018-06-29). `pool_pre_ping`
  has never been set.
- The catch-all was added **commented out** in `5a9530bd2` (2025-10-30,
  "harmonize api errors"), carrying the note
  `TODO: why cant' we handle all other errors here? (tests failing)`, and was
  **enabled** in `327a0523f` (2025-12-23, "api refactoring"). Before that a dead
  connection escaped as a 500 with a traceback; since then it is a quiet 400
  that reads like a bad request.
- Production now runs 4 mod_wsgi processes (measured 2026-08-11), so there are
  four independent pools. The same traffic spread over four pools means each
  pooled connection is reused less often and sits idle longer between uses.

Checked and **not** the cause: the bulk upload's session timeouts use
`SET LOCAL` (`api/actions.py:2172-2180`), so they are transaction-scoped and do
not follow the connection back into the pool.

## Fix

`pool_pre_ping=True` on the OEDB engine, plus a regression test that kills the
pooled backend and asserts the next request still answers 200.

jh-RLI:
> [!caution] Corrected 2026-09-21 after reading the production error log — **this does not fix the reported symptom**
> Counted over the whole per-vhost log history: `server closed the connection unexpectedly`
> appears **3** times; `remaining connection slots are reserved for non-replication superuser
> connections` appears **7,041** times. Both render as the same `400 {"reason": "Invalid request"}`
> through the same catch-all.
>
> So the fault fixed here is real and worth fixing, and it is **not** what the table data view has
> been failing on. That is connection-slot exhaustion on `oedb`, tracked as #2495. Do not read this
> PR landing as "the table view is fixed".
>
> The second commit is what earned its place: the catch-all logged `str(exc)` and nothing else,
> which is why two unrelated faults arrived as one anonymous 400.

Closes #2488.

Two commits, and the second one changes no response.

## `pool_pre_ping=True` on the OEDB engine

`{"reason": "Invalid request"}` is not a message about the query — it is the
catch-all in `api/helper.py`. The exception behind it was a dead pooled
connection: `pool_recycle` discards a connection for being **old**, never for
being **dead**, and `pool_pre_ping` was never set, so a connection Postgres or
the network had already closed was handed to the next checkout. SQLAlchemy
invalidated it on the way out, which is why the retry worked — and why the
whole thing read as a flaky table page rather than as a defect.

Reproduced and fixed against a local instance: warm the pool with one search,
`pg_terminate_backend` the idle backend, search again.

| | before | after |
|---|---|---|
| request after the kill | `400 {"reason": "Invalid request"}` | `200` |
| request after that | `200` | `200` |

Cost is one lightweight round trip per checkout.

The table data view is where you notice it because
`dataedit/static/dataedit/backend.js:468` fires the count query and the select
query **in parallel** for every draw: two checkouts at once, and either one
failing fails the draw.

## Why now

The fault is old and the way it presents is new. The **frequency** I have not
measured, and the first version of this section guessed at it and got it wrong
— the correction is below, because it is the more interesting half.

- Pool settings unchanged since `1a1bf52c2` (2018-06-29). `pool_pre_ping` has
  never been set.
- The catch-all was added **commented out** in `5a9530bd2` (2025-10-30,
  *harmonize api errors*) carrying the note
  `TODO: why cant' we handle all other errors here? (tests failing)`, and was
  **enabled** in `327a0523f` (2025-12-23, *api refactoring*). Before that a
  dead connection escaped as a 500 with a traceback; since then it is a quiet
  400 that reads like a bad request. **This is the only part of "why now" that
  is established.**

> [!warning] Correction
> An earlier version of this section said production runs 4 mod_wsgi processes,
> so there are four independent pools and each connection is reused less often.
> That is **wrong for this endpoint**. WF-03 of the DevOps 504 map (applied
> 2026-08-10) put `/api/v0/advanced` on its own `${APP_NAME}_api` group at
> **`processes=1 threads=15`**, precisely because `_SESSION_CONTEXTS` has no
> process affinity. Advanced traffic shares **one** pool, not four. The 4
> processes are the main group.

What actually bounds the window: `pool_recycle=600` discards a connection older
than 600 s **at checkout**, so only a connection killed within ~10 minutes of
being *created* can be served dead. What kills them on that host I have not
measured. The standing candidate is already on record — the OEDB pool is
unbounded and leaks (~3 `idle in transaction` a day, 12 observed live, DevOps
map WF-06), so it holds more connections than it turns over, which is the
condition for one to sit unused long enough to be reaped.

Checked and **not** the cause: the bulk upload's session timeouts use
`SET LOCAL` (`api/actions.py:2172-2180`), so they are transaction-scoped and do
not follow the connection back into the pool. That was done right.

## Making the next one diagnosable

The second commit changes what the catch-all logs, not what it returns. It
logged `str(exc)` — no exception type, no traceback, no path — so a dead
connection and a genuinely malformed query produced the same line and the same
body. That is most of why this survived: the production log could not identify
it.

The body stays generic on purpose and a test pins that it still leaks nothing.

## Note on the regression test

It disposes the pool first and kills exactly the one backend left in it, rather
than every backend on the database. The OEDB is a real local database in this
setup, not a throwaway test one, so a blanket `pg_terminate_backend` would take
out a second test run on the same machine.

## Testing

- `api`: 199 green
- full suite: 683 green without Fuseki, plus `oekg` 570 green against a local
  Fuseki — 1253 total
- `api.tests.test_openapi_schema`: 6 green (no route or serializer changed)
- both new behaviour tests confirmed red before the change

## Two things found alongside, not fixed here

Neither belongs in a hotfix; say the word and I will file them.

- **The search path commits a connection after returning it to the pool.**
  `api/helper.py:146-151` builds the triggers as
  `[close_cursor, close_raw_connection, connection.commit]`. `connection.commit`
  binds to the raw psycopg2 connection *before* `close_raw_connection` returns
  it to the pool, so the commit succeeds — against a connection another thread
  may already have checked out. Verified: the pool hands the same raw connection
  straight back out. Harmless for reads, less so for the insert/update paths
  sharing this decorator.
- **The table page can 400 itself under its own concurrency.**
  `USER_CONNECTION_LIMIT = 4` and every artificial-connection search consumes
  one, so parallel searches from one account reach
  `{"reason": "This user exceeded the connection limit..."}`. Different message,
  same endpoint.


jh-RLI:
Host capture on 2026-09-21 found a **second live cause of the identical 400**, so please read [this comment on #2488](https://github.com/OpenEnergyPlatform/oeplatform/issues/2488#issuecomment-5757409395) before merging. Short version: `oedb` is out of ordinary connection slots right now, that produces the same `{"reason": "Invalid request"}` body through the same catch-all, and `pool_pre_ping` cannot help while slots are exhausted — it detects the dead connection and then fails to open a replacement.

The fix here is still correct and still worth landing; it is **necessary, not sufficient**, and I have stopped short of claiming it is what the reporter was hitting. The real remedy for the exhaustion is bounding the pool (WF-06 on the DevOps 504 map), now the top item there.

The likely accelerant is also worth recording against this branch: the main mod_wsgi group went from `processes=4` to `processes=12` on or after 2026-08-21, and each process carries its own unbounded pool (`pool_size=0, max_overflow=200`).

Issue #2467, opened 2026-09-15
Delete a whole scenario bundle, in two steps


jh-RLI:
## Description of the issue

A scenario bundle can be created, read, changed and taken apart through the REST API
(#2434, #2439, #2443, #2451, #2455) but the bundle itself cannot be removed. A pipeline
that created a bundle it did not mean to has no way to take it back, and the browser's own
delete cannot be borrowed — it is the thing the previous slice exists to avoid repeating.

This is the ninth of the twelve OEKG REST API slices, from the wayfinder map's tickets note
(*"Delete a whole bundle, in two steps"*). It is unblocked by #2455 and is one of the three
strands *"Replace the whole bundle in one call"* collects.

**This is the one irreversible operation in the API and the single most dangerous thing it
can do**, so it is made deliberately awkward. Two guards, defending two different accidents:

- **the version** catches *somebody changed this since you looked*;
- **the retyped acronym** catches *right verb, wrong identifier* — the realistic failure for
  a pipeline looping over a list of identifiers, and the one a version cannot catch, because
  the version it names is the correct current version of the wrong bundle.

Three things today's delete path gets wrong that this one must not inherit:

- It walks one predicate one hop, so it **destroys shared model and framework nodes** and
  **orphans everything two hops out**. #2455 already fixed that bound for a part; a bundle
  delete is the same walk rooted one level up.
- It removes only the bundle's **outgoing** triples, so the version node — which points *at*
  the bundle — is left behind, asserting that a bundle nobody can read is at version 4. That
  residue is #2440.
- The **ownership rows outlive their bundles**: `ScenarioBundleAccessControl` is never
  cleaned up, so the table accumulates rows naming bundles that do not exist.

### 1. Two steps, and the first one is free

Step one is the read a client needs anyway: `GET` returns the version (as `ETag` and in
`_meta`) and the acronym. Step two is
`DELETE /api/v0/scenario-bundles/<uid>/?confirm=<acronym>` carrying `If-Match`.

The acronym is compared **exactly**. Normalising it would let `api-test` confirm a delete of
`API-TEST`, which is the confusion the check exists to catch.

### 2. Existence is checked first

Before authorisation, before the precondition, before the confirmation. A repeated delete
answers **404**, and the contract states a client may treat that as success after a lost
response.

A blanket success would swallow the wrong-identifier delete: a bundle that is gone has no
acronym left to check a confirmation against, so `204`-to-anything would confirm anything.

[...]

Issue #2461, opened 2026-09-15
Retire the legacy scenario-dataset documentation page


jh-RLI:
## Parent

Spec: #2454 — *Document and teach the OEKG REST API*. Derived from WF-14 decision D3, carried forward as spec decision D10 and note F1.

## What to build

The documentation stops carrying a page for a route the API has replaced, and stops promising **group-based shared editing**, which was ruled out of scope and does not exist. The databus registration prose on that page is the only explanation of databus registration anywhere in the documentation, so it moves rather than being lost.

**Why this is its own ticket:** the same deletion was placed on the *Study reports and dataset links* pull request, which merged without it. The page is still published, still promising the feature, and its navigation entry is still indented one level too shallow — which is why it renders as a sibling of whole sections rather than as a child of the OEKG API. The condition that deferred it (the route not yet superseded) is now met.

## Acceptance criteria

- [ ] The legacy scenario-dataset page is deleted.
- [ ] Its navigation entry is removed; the stray indent goes with it, so nothing is left rendering at the wrong level.
- [ ] The databus and external-URL prose is migrated into the client rules page.
- [ ] No remaining link in the documentation points at the deleted page.
- [ ] Changelog entry.

## Blocked by

- #2460 — the client rules page — the databus prose migrates into that page, so it must exist.

Issue #2454, opened 2026-09-14
Document and teach the OEKG REST API: generated reference, prose in two homes, two academy notebooks


jh-RLI:
> **Spec produced by `/to-spec` from the closed documentation half of the *Wayfinder Map — OEKG REST API from the shape* (tickets WF-13, WF-14, WF-15, all resolved 2026-09-14). The map is now closed.** The map, its fifteen tickets and this spec's sibling (the API contract, WF-01…WF-12) live in the maintainer's vault; this issue is the paste of the spec itself so the work is trackable here.
>
> The repository has no `ready-for-agent` label, so this carries `documentation` and `OEKG-API` instead — add the label if you want the convention from `CLAUDE.md` to be real.

---

# Spec — OEKG REST API documentation

Produced by `/to-spec` from the documentation half of
*Map - OEKG REST API from the shape* (WF-13, WF-14, WF-15), which closed on 2026-09-14 and closed
the map with it. Each decision below is derived from a resolved ticket; the ticket holds the
measurements and the rejected alternatives.

**The second of two specs.** *Spec - OEKG REST API from the shape* covers the API contract itself
(WF-01 … WF-12) and is being implemented as the twelve slices in
*Tickets - OEKG REST API from the shape*. This one covers how that API is **described and taught**
— the destination the maintainer added on 2026-09-08: *an API nobody can find their way into is not
finished.*

**Verified against the working tree on 2026-09-14**, after PR #2398 (drf-spectacular) and PR #2452
(slice 7) merged. Three of WF-13's and WF-14's stated facts had moved and are corrected in place
below.

---

## Problem Statement

The OEKG REST API is being built slice by slice, and **nobody outside the team can find their way
into it.**

A client author has no reference. The platform's OpenAPI description of its REST API is
**hand-written and 60 KB**, documents **47 paths against 65 routes**, and has already lost the
Dataset CRUD (#1971, merged), Bulk Upload (shipped in v1.9.0) and **everything OEKG**. The one
OpenAPI file that does mention the OEKG describes the SPARQL passthrough, not the CRUD surface.
So the experiment between hand-written and generated has already been run in this repository, and
hand-written lost.

A reader who does find a page finds a **navigation that has come apart**. The scenario-bundle
material is split across two plausible homes and one page escaped both: it is indented one level too
shallow and renders as a sibling of *Web-API's* and *Django Apps* rather than as a child of *OEKG
API*. That same page documents a route the API replaces, while promising group-based shared editing
[...]

Issue #2477, opened 2026-09-15
Replace the whole bundle in one call


jh-RLI:
## Description of the issue

The **twelfth and last** of the OEKG REST API slices, from the wayfinder map's tickets note
(*"Replace the whole bundle in one call"*), and the one the whole effort started from: a
modelling pipeline declares what the bundle should look like and the server makes it so.

Eleven slices are on `develop` (#2434, #2439, #2442, #2443, #2446, #2448, #2451, #2455,
#2467, #2471, #2468). This one collects the three that landed on 2026-09-15 — *Delete a whole
bundle, in two steps* (#2467), *The read side, completed* (#2471) and *Say whether a linked
dataset is still there* (#2468).

## What to build

A **separately named** endpoint — `POST /api/v0/scenario-bundles/<uid>/replace/` — taking the
complete desired bundle including nested sub-resources, applying the difference as one atomic
update.

Full replacement is deliberately **not** offered on the bundle's own URL. `PUT` was rejected
early for accident-proneness rather than for its semantics, and an endpoint you have to name
and guard is not hit by accident. Single-value edits stay on `PATCH`; the full payload is the
price of the declarative endpoint alone.

**Delete-by-omission is the semantics of this one endpoint and exists nowhere else in the
API.** Anything in the graph and absent from the payload is removed through the typed
containment walk of #2455, so shared nodes are still only unlinked, never deleted.

## Acceptance criteria

- [ ] `POST .../<uid>/replace/` takes the complete desired bundle, nested sub-resources included.
- [ ] It requires `If-Match`, with the same refusal order every other write uses: existence,
      ownership, precondition. `428` absent, `412` stale, `409` when the bound guard fails.
- [ ] Omission removes, through `removal.plan_removal`'s walk — so a node another bundle cites
      is unlinked rather than deleted, and the guard can still only downgrade.
- [ ] The response **reports what was removed and what was unlinked**, so a client asserts on
      the consequence rather than inferring it.
- [ ] A nested sub-resource carrying an identifier is matched to the existing one; without an
      identifier it is created. Without this, a re-import deletes and recreates every scenario
      on every run — churning identifiers and filling the history with false deletions.
- [ ] The whole operation is **one** request to the store and produces **one** history entry
      with a real diff. Never a delete followed by a create.
- [ ] The full pipeline story works end to end: look up by acronym (`GET ?acronym=`, #2471);
      create if absent; otherwise replace. **The client holds no state between runs.**
[...]

PR #2498, opened 2026-09-21
Point the pinning tests at the documentation pages they describe (#2462)

### LINKED ISSUE #2462: Point the pinning tests at the documentation pages they describe
## Parent

Spec: #2454 — *Document and teach the OEKG REST API*. Derived from WF-14 decision D9 and the third test in the spec's Testing Decisions.

## What to build

A test that pins a documented behaviour **names the page it describes**, in its docstring. The pointer runs from the code to the doc, not the reverse: a citation in prose rots silently on a rename, while a pointer in the test fires at the moment someone changes the behaviour — the only moment the page is wrong.

This is not a new habit. The mechanism already exists in this repository in exactly one place, and that place is instructive: the one test that points at a document points at a **research note in the maintainer's private vault**, which an external contributor cannot read — and it was written by the very migration that left a developer-guide table stating two dropdown lists were hardcoded after both had been made dynamic. Wrong about two fields at once, found by a reader rather than by a check.

> **The rule:** a test that pins a behaviour a documentation page describes names **that page** in its docstring — the page, not the vault note.

## Acceptance criteria

- [ ] Each existing pinning test for the rules that have one names the client rules page in its docstring.
- [ ] The sector-dropdowns test's pointer is corrected from the vault note to the scenario-bundles architecture guide.
- [ ] A test in the schema test module collects every code→doc pointer and asserts that each named documentation page exists in the tree.
- [ ] That test asserts what is present; it does not require a pointer for a behaviour that has not been built.
- [ ] Changelog entry.

## Carried obligation — not closable here

Two of the nine rules have **no pinning test yet**, because the behaviour is not built:

- rule 3 (delete-by-omission exists on `replace/` and nowhere else) → its pointer lands with *Replace the whole bundle in one call*
- rule 5 (the delete's two guards cover two different accidents) → its pointer lands with *Delete a whole bundle, in two steps*

Those two API slices should each gain the pointer as an acceptance criterion. Recorded here so the obligation is visible rather than assumed.

## Blocked by

- #2457 — the generated API reference — the pointer-integrity test lives in the test module that ticket creates.
- #2460 — the client rules page — the pointers name that page.
- #2458 — the API write path section — the corrected sector-dropdowns pointer names the page that ticket edits.

jh-RLI:
Closes #2462. Doc slice 6 of 9 under #2454.

## What this does

A documentation page describing behaviour goes stale in silence — nothing fails
at the moment the behaviour changes, which is the only moment the page is
wrong. A citation written into the prose does not help: it rots on a rename
with nobody watching. So the pointer runs the other way. The test names the
page, in its docstring, where the next person to change the behaviour is
already reading.

**Eleven tests carry a pointer, covering all nine rules of the client guide**,
plus the architecture guide and the reference. `DocumentationPointerTest` in
`api/tests/test_openapi_schema.py` collects them and fails when a named page is
not in the tree.

## The mechanism, and why it is shaped this way

A pointer is one line in any docstring: `Documents: <repo-relative path>.md`.
The prose around it is for the reader; that line is the part a machine reads,
so the sentence a developer sees and the fact the check collects are one thing
rather than two that can drift.

Three properties are deliberate:

- **It reads the source, it does not import it.** Importing would be far worse
  than slow: `factsheet/tests/test_sector_dropdowns.py` imports
  `factsheet.oekg.connection`, which parses the whole OEO at module scope
  (`factsheet/oekg/connection.py:40`), so collecting docstrings by import would
  pull a 1.3 GB ontology into memory to read a docstring.
- **It scans this project's own apps, not the tree below `BASE_DIR`.** Measured,
  not assumed: a checkout routinely holds its virtualenv inside the repository,
  and this one has **1,922 files matching `test*.py` under `.venv`**. Being
  under `BASE_DIR` does not separate ours from theirs either — with a venv in
  the checkout, all **27** installed third-party apps are under it too. The app
  registry plus the location of `site-packages` does. A `Documents:` line in any
  dependency's docstring would otherwise be reported as this project's broken
  pointer.
- **It asserts what is present and demands nothing.** A check that required a
  pointer would go red until unbuilt work landed, which teaches readers to
  ignore a red suite.

**The class carries no pointer itself, on purpose.** The first version did, and
it made the emptiness guard satisfy itself out of its own docstring — the check
would then have passed on a suite from which every real pointer had been
deleted. Both failure modes are exercised: I temporarily pointed a test at a
renamed page and confirmed the check names the file and the missing page.

## The pointer that already existed named a private vault

`test_sector_dropdowns.py` cited *"the OEKG scenario-bundles wayfinder research
(WF-01 and its NC/BR addendum)"* — a note an external contributor cannot open.
It names `features/scenario-bundles/architecture.md` now. That is the page whose
table stood wrong about two fields, after both were made dynamic, until a reader
happened to open it — which is the drift this whole ticket exists for.

## Second half: the citation guard read one page

Carried over from reviewing #2496, which merged this morning and was flagged on
both review axes as its sharpest finding. `test_api_reference_tags.py` checked
the reference's own section links, while **two other pages had come to cite
`Scenario Bundles (legacy)`**, so renaming that group would have gone red in one
place and quietly dead in two. It reads every documentation page now.

**It was not the two-line change it looked like.** Extending it surfaced that
the old pattern also mis-parsed a citation ending a Markdown link, taking the
link's own closing bracket for part of the section name (`Scenario Bundles)`) —
invisible while the single page checked happened to write its one citation in
backticks. Parentheses are part of a group's name, so they are now matched only
in balanced pairs.

## Two corrections to the issue body

1. **The carried obligation is fully discharged, not partly.** The issue defers
   both rule 3 and rule 5. Rule 5 was already wrong when the ticket was written —
   *Delete a whole bundle, in two steps* merged on 2026-09-15 (#2467 / PR #2470).
   And **#2494 merged while this branch was being written**, so rule 3's
   behaviour is built too. `test_bundle_replace.py` pins exactly what rule 3
   claims, so it gets the pointer here.
2. The issue's *"its pointer lands with Replace the whole bundle in one call"* —
   that slice landed without it, the same way #2461's page deletion rode a pull
   request that merged without it.

## One thing found and deliberately NOT fixed

**#2494 shipped `replace/` but left the client guide saying it does not exist.**
Rule 3 still carries a "This endpoint is not built yet" warning, a `404` where
the worked example belongs, and the line "In the reference: nothing yet". None
of that is true any more.

I have not repaired it, and the reason is the page's own contract: it states
that every exchange on it is real and recorded against a running graph store.
Writing the worked example it is owed needs a store, not a guess — so this is a
follow-up, either as an amendment to #2494 or its own ticket. Say the word and
I will file it.

Worth noting what this means for the mechanism: a pointer names a page, so the
check catches a page that is *gone*, not one that is *wrong*. It is not a claim
to have solved staleness, and this is a clean example of the remaining gap.

## Checks

- Full suite, in a test database of its own: **796 tests, `OK (skipped=101)`**
  (1,336 found; the skips are the OEKG graph tests, which need a Fuseki).
- `api.tests.test_openapi_schema` + `api.tests.test_api_reference_tags` — 22
  green.
- `reuse lint` — compliant, 1206/1206.
- Merged `origin/develop` in (it moved twice during the work) and re-ran
  everything. The merge was clean: git took both sides of
  `oekg/tests/test_dataset_link_api.py`, which #2494 and this branch both touch,
  and the docstring pointer survived — verified by hand, not assumed.

## Review note

`/code-review`'s two sub-agents both died on a session rate limit, so I reviewed
the diff myself against the same two axes and applied three fixes: the parse no
longer swallows exceptions (scoping to our own apps removed the reason for it,
and a file of ours that will not parse should be loud); `INSTALLED` became
`SITE_PACKAGES`, because beside Django's `INSTALLED_APPS` the old name read as
its opposite; and a sentence noting the guide is currently out of date came back
out of a docstring — a note that something is wrong *today* is exactly the prose
this convention exists not to rely on. Worth a human eye on the two regexes in
particular.


jh-RLI:
Filed the finding from *One thing found and deliberately NOT fixed* as **#2500**, and added it to the plan as a sub-issue of #2454.

It turned out to be worse than the one stale sentence I described here: **seven** false claims on the page, not one. The sharpest is line 243 — *"In the reference: nothing yet"* — which is false against an artifact in this same repository, two directories away: `openapi.yaml` already carries `scenario_bundles_replace_create`, tagged *Scenario Bundles*, because the drift guard regenerated it when #2494 merged. Line 240 also points at **#2477 as the tracking issue for work still to come**, and that issue is closed.

The reason for not repairing it here stands and is written into #2500: the page states that every exchange on it is real and recorded against a running store, so rule 3's worked example needs a Fuseki rather than a careful guess.

Worth recording as a pattern rather than an incident — this is the **third** time on spec #2454 that an obligation placed on a feature slice's pull request merged without it: #2461's page deletion, rule 5's pointer, and now rule 3's prose. #2460 shipped the page with rule 3 marked pending on the explicit understanding that "the page is edited a second time when `replace/` lands".

Issue #2473, opened 2026-09-15
Replace must not delete dataset links by omission: they are bundle-local but not in the bundle payload


jh-RLI:
## Description of the issue

**Blocks the replace slice.** Not a defect in anything merged — a decision the replace
endpoint cannot be built without, and one that fails silently in the direction of data
loss if it is missed.

Dataset links are **not part of the bundle payload**. A bundle read nests scenarios and
study reports and nothing else — `_represent` loops `BUNDLE_PARTS` (`oekg/api_views.py`),
and a dataset link is not a part; a nested bundle `POST` builds the same set
(`build_bundle_graph`, `oekg/bundles.py`). Links live only under
`.../scenarios/<sid>/datasets/`.

But their classes **are** bundle-local for the purposes of deletion:

```python
# oekg/removal.py
BUNDLE_LOCAL_CLASSES = frozenset({
    SCENARIO_CLASS,                          # scenario factsheet
    STUDY_REPORT_CLASS,                      # study report
    DIRECTION_BY_NAME["input"].node_class,   # input dataset link
    DIRECTION_BY_NAME["output"].node_class,  # output dataset link
})
```

That is correct for #2456 and for the whole-bundle delete (#2470): a scenario's links
should go with the scenario rather than be orphaned.

**It is not correct for delete-by-omission.** The replace slice's contract is *"Anything in
the graph and absent from the payload is removed through the typed containment walk"*, and
the payload it is given has no way to mention a dataset link. So a replace built on today's
bundle shape would **delete every dataset link in the bundle**, on every run, with the
client having had no way to prevent it.

The trap is written into the code as an assertion of the opposite. `_represent`'s docstring
says:

> a read sent back to `POST` copies the whole bundle, scenarios and study reports included,
> and **later the replace endpoint takes exactly this shape**

Taking exactly this shape is the bug. And the happy-path test one would naturally write for
replace — add a scenario, change a field, remove a scenario — passes while citations are
destroyed.

Verified against `origin/develop` and against all three open API branches: #2469
(resolution), #2470 (bundle delete) and #2472 (read side). None of them changes either
half; #2472 leaves `_represent` looping `BUNDLE_PARTS`.

## Ideas of solution

Two questions, and they are one decision.

### 1. Do dataset links join the bundle payload?

The natural home is nested under each scenario, on read and on the nested create, matching
how scenarios and study reports already nest under the bundle. That restores the property
[...]

Issue #2460, opened 2026-09-15
The client page: the nine semantic rules of the scenario-bundle API


jh-RLI:
## Parent

Spec: #2454 — *Document and teach the OEKG REST API*. Derived from WF-14 decisions D5 and D8.

## What to build

A client author reads **one page** and learns the behaviour a signature cannot express — the rules they would otherwise discover by hitting a `400`. Everything mechanical (`If-Match`, `?confirm=`, `?expand=`, the `428`/`412`/`409` table, RDF negotiation) stays in the reference and is **not** restated here.

The nine rules:

1. A key absent from a payload means **untouched**, under every verb.
2. `_meta` is read-only and ignored on write.
3. Delete-by-omission exists on `replace/` and **nowhere else**.
4. A bundle `POST` accepts nested sub-resources; a `PATCH` does not.
5. The delete's two guards cover two **different** accidents — stale state and wrong uuid — so neither is redundant ceremony.
6. A write is judged by the violations it **introduces**; a create is strict because it has no pre-state. A client seeing `400` on create where a `PATCH` succeeded otherwise reads it as a bug.
7. `ref: table` is reproducible, `ref: dataset` is current; a dead link **stays** and reads `resolvable: false`, which is correct rather than broken.
8. The **server** mints the uid; a stateless pipeline re-identifies with `GET ?acronym=`, which also yields the `ETag`.
9. `_meta.history_recorded` can read `false` inside a **successful** write — a named audit gap, not a status flag.

## Acceptance criteria

- [ ] A new page under the OEKG API section carries all nine rules.
- [ ] Each rule has a worked request and response example.
- [ ] Each rule links the reference at its tag anchor rather than restating the endpoint table.
- [ ] The page does not repeat mechanical facts that the reference already carries as annotations.
- [ ] Changelog entry.

## Blocked by

- #2459 — the endpoint-group tags — the rules link by tag anchor, so the anchors must be stable and singular first.

**And by the API itself.** Checked against the routes on `develop` (2026-09-15): four of the nine rules describe behaviour that does not exist yet, so this page cannot ship honestly before those slices land. The endpoints present today are the collection `POST`, the detail `GET`/`PATCH`, history, scenarios, study reports and dataset links — there is no `DELETE`, no `replace/`, no `?acronym=` and no link resolution.

| rule | state today |
|---|---|
| 1, 2, 4, 6, 9 | built — documentable now |
[...]

PR #2502, opened 2026-09-21
Rule 3 of the client guide describes replace/ as it is

### LINKED ISSUE #2500: The client guide still says `replace/` is unbuilt, and it shipped
## Parent

Spec: #2454 — *Document and teach the OEKG REST API*. Found while building #2462
(PR #2498) and reported rather than repaired there; see *Why this is its own
ticket* below.

## What is wrong

**PR #2494 shipped `POST /api/v0/scenario-bundles/{uid}/replace/`. The client
guide still tells readers it does not exist.**

The endpoint is real and measured, not assumed:

- it is routed — `api/urls.py:344`, `scenario-bundles/<uid>/replace/` →
  `ScenarioBundleReplaceAPIView`;
- it is **already in the committed reference** —
  `docs/oeplatform-code/web-api/openapi.yaml` carries
  `/api/v0/scenario-bundles/{uid}/replace/`, `post`, operationId
  `scenario_bundles_replace_create`, tagged *Scenario Bundles*;
- it has 918 lines of tests — `oekg/tests/test_bundle_replace.py`.

Seven claims on `docs/oeplatform-code/web-api/oekg-api/scenario-bundles.md` are
now false. The last one is the sharpest, because the document it denies is
generated from this same repository and sits two directories away:

| line | claim |
|---|---|
| 20–21 | "Eight describe the API as it is today; **rule 3** waits on an endpoint that is still being built" |
| 208 | heading: "delete-by-omission **will exist** on `replace/`" (future tense) |
| 210 | admonition title: "This endpoint is not built yet" |
| 212 | "`POST /api/v0/scenario-bundles/{uid}/replace/` does not exist on the platform today" |
| 219–228 | a worked exchange whose response is `404 Not Found` |
| 241 | "This page gains the worked example when it lands." |
| 243 | "In the reference: nothing yet." |

The rule's body is future-tense throughout ("it **will** take the complete
bundle you mean to see"), and line 240 points at **#2477 as the tracking issue
for work still to come** — that issue is closed.

## Why this matters more than a stale sentence

Rule 3 is the one rule on that page about **destroying data**. The page's whole
argument is that leaving a key out is safe everywhere, and that this stops being
true at exactly one address. A reader is currently told that address is not
reachable yet. It is.

## Why this is its own ticket

#2498 is the code→doc pointer slice, and it adds a pointer from
`test_bundle_replace.py` to this page. It deliberately did **not** repair the
prose: the page states that *"every exchange below is a real one, recorded from
the API running against a real graph store, and the responses are reproduced as
they came back."* Writing the worked example rule 3 is owed therefore needs a
running Fuseki to record against — it cannot be written from the code by
inspection without breaking the page's own contract with its readers.

This is also a clean illustration of the limit of #2462's mechanism, worth
keeping in mind: a pointer names a page, so the check catches a page that is
**gone**, not one that is **wrong**.

## Acceptance criteria

- [ ] Rule 3 describes `replace/` as it is, in the present tense: the
      admonition, the heading, the intro sentence at line 20 and the "in the
      reference" line all go.
- [ ] The `404` exchange is replaced by a real recorded one, against a running
      store, per the page's stated convention — a request that replaces a bundle
      and a response showing what was deleted and what was unlinked.
- [ ] The rule states the property `ReplaceKeepsDatasetLinksTest` exists for:
      a replace of a bundle carrying dataset links does not destroy them
      (the #2473 hazard).
- [ ] The reference link points at the operation, which already exists:
      `scenario_bundles_replace_create` in the *Scenario Bundles* section.
- [ ] The stale reference to the closed #2477 goes.
- [ ] Changelog entry.

## Note on how this happened

The same shape as #2461: an obligation placed on a feature slice's pull request,
which merged without it. #2460 shipped the page with rule 3 marked pending on
the explicit understanding that "the page is edited a second time when
`replace/` lands". `replace/` landed; the second edit did not.

jh-RLI:
## Summary of the discussion

#2494 shipped `POST /api/v0/scenario-bundles/{uid}/replace/`. The client guide
went on telling readers it did not exist: a warning admonition titled *"This
endpoint is not built yet"*, a worked exchange answering `404`, the future
tense throughout, and a pointer to #2477, which is closed. All seven claims
listed in #2500 are gone.

Rule 3 is the only rule on that page about a write that destroys data nobody
named, so the address where omission removes things was documented as
unreachable.

### The exchange is recorded, not written

The page promises its readers that *"every exchange below is a real one,
recorded from the API running against a real graph store, and the responses are
reproduced as they came back."* That promise is why #2498 reported this rather
than repairing it: the worked example rule 3 is owed needs a running Fuseki.

Against a local instance: a bundle was created with one scenario carrying two
dataset links (one `ref: table`, one `ref: external`) and one study report, read
back, and declared afresh with the abstract changed, a second scenario added,
and **no `study_reports` key at all**. The answer is `200`, `ETag: "3"`, with
`_meta.deleted` naming the study report and both citations surviving with the
identifiers they were read with. A script checks every value printed on the page
against the raw recording; the only edit is the host name, which the page
already says it rewrites.

The idempotence sentence is recorded too: declaring the same bundle again left
the entity tag at `"3"` and the history at 3 entries.

### The #2473 property is stated

The rule now says the thing `ReplaceKeepsDatasetLinksTest` exists for. Dataset
links nest under their scenario in the bundle payload and are bundle-local for
the purposes of deletion, so a declaration that carries a scenario but not its
`datasets` deletes that scenario's citations — and the happy path anybody would
write by hand (add a scenario, change a field, remove a scenario) does not
notice. That warning replaces the one that said the endpoint was unbuilt.

### Two things beyond the ticket, both flagged for a nod

**Rule 2 was contradicting rule 3.** It stated flatly that `_meta` is
read-only and dropped before validation, which has not been true since #2494:
`replace/` reads `_meta.uid` to tell which nested resource you mean. Putting
rule 3 on the same page in the present tense made the page contradict itself,
so rule 2 now names the one exception and the two rules point at each other —
the page's convention for a pair that qualifies each other. One clause, but it
edits a rule the ticket did not mention.

**The release notes contradict themselves and I have not touched that.**
`versions/changelogs/current.md` still carries #2460's entry saying rule 3
*"describes an endpoint that does not exist yet and is marked as pending"* and
#2462's *"a rule whose endpoint is still being built"*. Both were accurate when
written. Rewriting another PR's entry is retroactive, so this is reported
rather than done — say the word and it is two small edits.

### Review

`/code-review` ran on the first commit and found three things, all fixed in
`921c0d338`:

- **the worked example demonstrated the wrong half of the rule.** It sent
  `"study_reports": []`, which is rule 1's explicit empty-list removal and works
  on a `PATCH` too — not omission. Re-recorded with the key genuinely absent;
  that the two are equivalent here is now said in prose instead of standing in
  for the example. (`oekg/replacement.py:173,213` is where they become the same.)
- rule 2's contradiction, above;
- the intro claimed rule 3 was the only rule about destroying data, which rule 5
  also is.

### Documentation updates

- Rule 3 of the client guide for scenario bundles describes `replace/` as it is
  [(#2500)](https://github.com/OpenEnergyPlatform/oeplatform/issues/2500)

### Automation

Closes #2500

### PR-Assignee

- [x] 📝 Update the
- [x] 📙 Update the documentation on

### Reviewer

- [ ] 🐙 Follow the

---

Checks run locally: `api.tests.test_openapi_schema`,
`api.tests.test_api_reference_tags` and
`api.tests.test_scenario_bundle_description` — 36 green, so the drift guard
passes and #2498's pinning pointer from `oekg/tests/test_bundle_replace.py`
still resolves. `mkdocs build` clean, and all four internal anchors the new text
uses resolve to headings that exist — worth doing by hand, because
`mkdocs build` runs only on push to develop, never on a PR.


jh-RLI:
Two follow-ups from review, pushed as `531da9738`.

**The `_meta.uid` question — the client cannot overwrite it, and the page now says so.**
The worry was fair and the page left it open. Measured against a running store rather than reasoned about:

- a nested sub-resource naming a `uid` this bundle does not hold is refused **400**, nothing written. `replacement._matched_part` looks the identifier up in the pre-state and raises; it never mints from a client's value.
- the bundle's own `_meta.uid` in the body is ignored outright. A declaration carrying a foreign uid *and* `version: 99` came back with the real identifier from the URL and left the entity tag where it was.

So the identifier is still the server's under every path. Rule 2's exception now says it is a **name, not an assignment**, and rule 3 carries that `400` as a second recorded exchange — it is the concrete thing a pipeline that assembled its payload from the wrong bundle meets.

That refusal's request block shows **headers only**, deliberately. The body was a complete valid declaration with one value swapped; printing an abbreviated one would have been an invented example on a page whose whole contract is that it has none, and a payload containing only that fragment would fail serializer validation first and produce a different `400`.

**The stale sibling changelog entries are dated.** #2460's said rule 3 *"describes an endpoint that does not exist yet"* and #2462's justified its check with *"a rule whose endpoint is still being built"*. Both were accurate when written and would have shipped in one release contradicting the entry above them.

Re-checked after both: `mkdocs build` clean, all five internal anchors resolve, both exchanges verified value-by-value against the raw recordings, `api.tests.test_openapi_schema` + `test_api_reference_tags` 22 green.

